PR #58, opened 2015-01-28
Shift-Reduce Constituency Parser

skystrife:
Please don't merge this just yet (I want to comment everything first), but I'm making the PR now so that you can do some code review if you'd like.

This is a rather large pull request (biggest one yet?) that contains all of the parser implementation, including the tree library (parse tree representation, visitors, transformers, PARSEVAL/evalb metrics) as well as the parser itself.

I've only tested training on the English Penn Treebank, bit it should be able to be modified to support other languages in the future. We would have to add classes like `head_finder` for different languages, but it should be doable.

Memory requirements are pretty light, surprisingly, and we seem to train faster than CoreNLP's shift-reduce equivalent.

Performance of my current model on the standard training/test split of the Penn Treebank:

```
Matched: 38531
Gold:    44276
Test:    44463
Labeled Recall:    87.0246
Labeled Precision: 86.6586
Labeled F1:        86.8412
Perfect Matching:  32.9056
Average Crossing:  1.29719
Zero crossing:     59.2715
```


include/parser/trees/parse_tree.h
```diff
@@ -0,0 +1,81 @@
...
+ * Represents the parse tree for a sentence. This may either be a sentence
+ * parsed from training data, or the output from a trained parser on test
+ * data.
+ *
+ * @todo determine what parts of analyzers::parse_tree are worth
+ * keeping---that class deals specifically with trees read from the output
+ * of the Stanford parser. When we have our own, we may still want some of
+ * that functionality to allow people to use parsers that are not our
+ * own?
+ */
```
smassung: The tree format that `analyzers::parse_tree` is reading is actually just the output of meta-stanford-preprocessor. So it could be output from any parser since the format is just nested parens. Non-parens are treated as tags/labels.

src/parser/io/ptb_reader.cpp
```diff
@@ -0,0 +1,138 @@
+/**
+ * @file ptb_reader.cpp
+ * @author Chase Geigle
```
smassung: This looks very similar to the `analyzers::parse_tree` string constructor. It looks like the only difference is that the PTB format allows newlines between characters. If that is true, we might be able to completely remove it after adding some very basic functions to the other `parse_tree` object. Separating the reader into this file is nicer anyway.

src/parser/state.cpp
```diff
@@ -0,0 +1,363 @@
...
+    return state.finalized();
+}
+}
+
+bool state::legal(const transition& trans) const
+{
+    // this is sort of ugly, but I think the ugliness introduced into the
+    // sr_parser class by inducing an inheritance hierarchy of transitions
+    // outweighs this class's ugly methods
+    switch (trans.type())
```
smassung: Agree

Issue #52, opened 2014-12-17
Streamline installation instructions


skystrife:
The top of the README/front page of the website should be modified to

1. [x] Contain concrete minimum system requirements for Ubuntu as well as OSX
2. [x] Contain instructions on installing the ICU dependency on the latest Ubuntu as well as OSX.
3. [x] Contain instructions on getting CMake >= 3.0 on Ubuntu as well as OSX
4. [x] Make it more immediately clear how to check out from git with all submodules (e.g. `git clone https://github.com/meta-toolkit/meta.git --recursive`)
5. [x] Make it more immediately clear how to choose between `g++` and `clang++` as a compiler (basically being Linux == `g++`, OSX == `clang++` unless you really know what you're doing)

"Favoritism" shown for Ubuntu/OSX because those seem to be the most common compatible platforms we're getting installation questions about. People not on those distros are probably more likely to be able to figure out how to install the dependencies based on the instructions given for Ubuntu (but we might eventually want to include a section for other distros...)

PR #65, opened 2015-02-03
Documentation Fixes

skystrife:
This PR addresses issue #52 by adding detailed setup guides for OS X, the two currently supported LTS versions of Ubuntu, and Arch Linux (because why not).

I tested everything on my local system using docker to get the different versions of Ubuntu, and booted my laptop into OS X v10.9 for the OS X instructions. They seem to be correct from those tests, and with the latest changes in develop I can easily run the unit tests on all three (everything passes).

I ended up not really saying much about selecting your compiler since the reality is that each system sort of has its own "best compiler" with the exception of Arch (for whom I doubt an install guide is even really necessary).

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -9,6 +9,7 @@ about MeTA!
 - develop: [![Build
   Status](https://travis-ci.org/meta-toolkit/meta.svg?branch=develop)](https://travis-ci.org/meta-toolkit/meta)
 
+### Outline
 - [Intro](#intro)
     - [Documentation](#documentation)
     - [Tutorials](#tutorials)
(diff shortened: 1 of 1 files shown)


skystrife:
Assigning @smassung mainly for proofreading and other comments/suggestions. Feel free to merge if you think it's ready, or suggest changes I can make before it's mergable.

skystrife:
Should I add a guide for EWS/EngrIT-managed systems, since we might expect some UIUC users?

smassung:
Yes, EWS guide would be great

skystrife:
Added.

smassung:
Can confirm EWS guide works on that system

PR #73, opened 2015-03-05
Updated ir_eval.cpp

husseinhazimeh:
Fixed the denominator of avg_p and made it equal to the number of relevant documents as calculated from the relevance judgements file.

### DIFF
--- a/src/index/eval/ir_eval.cpp
+++ b/src/index/eval/ir_eval.cpp
@@ -181,8 +181,8 @@ double ir_eval::avg_p(const std::vector<std::pair<doc_id, double>>& results,
             break;
     }
 
-    scores_.push_back(avgp / (i - 1.0));
-    return avgp / (i - 1.0);
+    scores_.push_back(avgp / ht->second.size());
+    return avgp / ht->second.size();
 }
 
 double ir_eval::map() const


smassung:
Thanks! The denominator for average precision should be the total number of relevant documents in the index. See http://en.wikipedia.org/wiki/Information_retrieval#Average_precision

smassung:
Oops, didn't merge the first time.

skystrife:
Hold up, I think this is more nuanced than that.

Isn't this function computing AveP@K? where K = `num_docs`? I'm thinking the denominator should actually be `std::min(ht->second.size(), num_docs)`...

skystrife:
(It could also be that I'm just confused since I've never looked at this code!)

smassung:
Hmm, good point. It is computing AveP@k, k = `num_docs`. According to the wiki page, the formula counts up from 1 to *n* documents, but the denominator is still all relevant documents.

But, if we returned the top 10 documents, and all 10 were relevant, then it would make sense that average precision is 1.0, correct? But what if there are 20 total relevant documents? Then this version would return 0.5.

So I'm thinking we go with the `min` version. Hussein, is this your opinion as well?

husseinhazimeh:
I'm not sure about this. According to
http://research.microsoft.com/pubs/57213/ecir2008.pdf (section 2.4) the
denominator should be equal always equal to the number of relevant
documents irrespective of the cutoff k. But I see your point.. by taking
the minimum, the value of the average precision will be more meaningful.
Whichever method you prefer, I think it's a good idea to document this
nuance in the code.

On Wed, Mar 4, 2015 at 8:45 PM, Sean Massung <notifications@github.com>
wrote:

> Hmm, good point. It is computing AveP@k, k = num_docs. According to the
> wiki page, the formula counts up from 1 to *n* documents, but the
> denominator is still all relevant documents.
>
> But, if we returned the top 10 documents, and all 10 were relevant, then
> it would make sense that average precision is 1.0, correct? But what if
> there are 20 total relevant documents? Then this version would return 0.5.
>
> So I'm thinking we go with the min version. Hussein, is this your opinion
> as well?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/meta-toolkit/meta/pull/73#issuecomment-77296403>.
>



-- 

[image: View Hussein Hazimeh's profile on LinkedIn]
<http://lb.linkedin.com/in/husseinhazimeh>

skystrife:
Interesting. Apparently we all were right! See: http://link.springer.com/referenceworkentry/10.1007%2F978-0-387-39940-9_487

However, I still think using `min(n, R)` makes the most sense, and that seems to be what TREC is doing.

husseinhazimeh:
Interesting indeed. Thanks for sharing this!

On Wed, Mar 4, 2015 at 9:45 PM, Chase Geigle <notifications@github.com>
wrote:

> Interesting. Apparently we all were right! See:
> http://link.springer.com/referenceworkentry/10.1007%2F978-0-387-39940-9_487
>
> However, I still think using min(n, R) makes the most sense, and that
> seems to be what TREC is doing.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/meta-toolkit/meta/pull/73#issuecomment-77301374>.
>



-- 

[image: View Hussein Hazimeh's profile on LinkedIn]
<http://lb.linkedin.com/in/husseinhazimeh>

PR #74, opened 2015-03-24
Add Docker image

zero323:
- add docker directory
- create simple Dockerfile using HEAD

I can tag/version specific images later.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# [v1.3.2][1.3.2]
+## Bug fixes
+- fix potentially incorrect generation of vocabulary map files on 32-bit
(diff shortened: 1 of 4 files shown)


skystrife:
This looks great. A few (really minor) things:

- I think we should put this in `contrib/docker`. 
- Could you re-make this against the `develop` branch? We tend to keep `master` as the latest released version.

Thanks!

smassung:
I'm not too familiar with Docker, but would there be any way to have it download the model files for the correct release? Just a suggestion!

zero323:
@skystrife Done. Please let me know if it is better now.
@smassung Only practical limitation is size of the image. If code changes are more frequent than the model files it can be better to keep data only images.

PR #75, opened 2015-03-25
Add Vagrantfile

31z4:


smassung:
Thanks, this should be useful for the MOOC!

PR #81, opened 2015-04-15
Document metadata

skystrife:
This branch adds support for generic metadata for documents, meaning we can now store arbitrary `uint64_t`, `int64_t`, `double`, and `string` typed information on a per-document basis. This makes adding things like boosts for specific documents, PageRank, etc. straightforward, and will support the development of a regression component of the toolkit (since we can now have a "label" that is a `double` instead of just a `class_label`). The one piece of document metadata that we are still storing the "old way" are the labels, since we need to do so many different things with those that I think it's OK to special case them.

This is a fairly large change to the corpus format and configuration, but nothing too terrible.

Configuring a corpus now looks like:

```toml
prefix = "/path/to/data/folder"
dataset = "ceeaus"
corpus = "line.toml"
```
where "line.toml" is a file at `prefix/ceeaus/line.toml` with contents like the following:

```toml
type = "line-corpus"
encoding = "shift_jis"
metadata = [{name = "grade-level", type = "double"},
            {name = "institution", type = "string"},
            {name = "student-id", type = "uint"},
            {name = "delta", type = "int"}]
```

The `metadata` key above describes the format of the additional metadata available for this corpus. This key is optional. If provided, the corpus will look for a file in the corpus directory called `metadata.dat`, which is expected to be a **TAB** separated file where each line denotes the metadata for documents in the corpus (from the first to the last).

The unit tests pass locally---in order to run them on Travis-CI, we will need @smassung to update the ceeaus corpus downloaded in the CMake files to include the additional files `line.toml`, `file.toml`, and `gz.toml` with the appropriate configurations:

```toml
# file: line.toml
type = "line-corpus"
encoding = "shift_jis"

# file: file.toml
type = "file-corpus"
list = "ceeaus"
encoding = "shift_jis"

# file: gz.toml
type = "gz-corpus"
encoding = "shift_jis"
```

I've also simplified the document representation in the case that we're reading them out from a `file_corpus`---that special case doesn't really make much sense to have around anymore, so documents now always float around with their full text in them, or have no content. We can revisit this if it's ever a problem (e.g., we have really stupidly huge documents).


include/corpus/metadata.h
```diff
@@ -0,0 +1,317 @@
...
+     */
+    template <class T>
+    util::optional<T> get(const std::string& name)
+    {
+        for (uint64_t i = 0; i < schema_.size(); ++i)
+        {
+            switch (schema_[i].type)
+            {
+                case field_type::SIGNED_INT:
+                    int64_t si;
```
smassung: I didn't know you can have cases without brackets for their content. Do you prefer to style it this way? Is there still a scope for each label?

skystrife: I actually don't know.

I think it's fine for integers, since there isn't any initialization that's being skipped, but I'm not sure why it doesn't complain later down when I construct a string, though... is it maybe because it's the last case and can't be jumped over? In either case, I can put braces around the cases to be more explicit if you want.

smassung:
Let me regenerate the necessary CEEAUS files and hashes and get the unit tests passing. Then I should be able to return and merge. This is looking like a 2.0 release when merged in master?

include/corpus/metadata.h
smassung: Ok, let's add the braces then just to be explicit about what's going on here.

skystrife:
It is certainly either a v1.4.0 or a v2.0.0.

include/corpus/metadata.h
smassung: And I'm sure you know there are two other places with `switch`

skystrife:
Once we get actual server space we should put the unit test datasets there and change the CMakeLists to point there, so then either of us can update the files.

smassung:
Oh, I definitely agree with that.

For the versioning, I think we have to decide if this merge makes "incompatible API changes". I think it's actually debatable, right? If clients update the config files for their datasets they shouldn't have to modify any code unless they want to use the metadata. It's fine if you need to regenerate indexes. The only other thing is that `forward_index` won't be a libsvm format (although you do provide the libsvm formatter). What do you think?

include/corpus/metadata.h
skystrife: The other ones aren't declaring variables, though, so I don't think it's necessary.

skystrife:
Well, the document class has changed (no longer mandates storing a name). The corpus classes have changed since they now have a metadata responsibility.

This is probably a v2.0.0.

Currently working on travis-ci, I'll ping you when I think this is mergeable.

smassung:
Ok cool. Do you think we should increase the timeouts for travis in debug mode? Or are the tweaks you're working on supposed to eliminate that need?

skystrife:
The travis-ci changes in here now should prevent us from having to mess with the timeouts.

If this build passes, I think we're ready to merge.

smassung:
Cool, it's using ninja now.

And g++ is still acting up apparently.

smassung:
`g++-4.8: internal compiler error: Killed (program cc1plus)`

...

skystrife:
-sigh- I have no idea wtf is going on there. Best guess I have is that we ran the container out of memory? (But this literally *just* worked in the other branch.) I've switched back to `make` to see if that helps.

skystrife:
That seems to have fixed it.

smassung:
Yay, merged.

PR #84, opened 2015-04-22
Ranking optimizations with fastapprox

smassung:
Changes:
- Use [fastapprox](https://code.google.com/p/fastapprox/) for fast log calculations in the rankers. I only included part of fastapprox in `include/util/fastapprox.h` since there are many warnings (some about GCC-specific features).
- Use `float` instead of `double` in ranking and scoring calculations
- Replace `std::pair<doc_id, float>` with `search_result` object for better readability and ease of use

With these changes, we get increases in ranking speed. Using BM25 (which uses one log calculation per term) on the `homepages` dataset and 500 TREC terabyte track queries improves query speed from 53 seconds to 44 seconds on my laptop.

We can also look into using fastapprox functions in other areas of MeTA, such as topic modeling or classification if necessary.


src/index/ranker/okapi_bm25.cpp
```diff
@@ -15,24 +16,24 @@ namespace index
...
-double okapi_bm25::score_one(const score_data& sd)
+float okapi_bm25::score_one(const score_data& sd)
 {
-    double doc_len = sd.idx.doc_size(sd.d_id);
+    float doc_len = sd.idx.doc_size(sd.d_id);
 
     // add 1.0 to the IDF to ensure that the result is positive
-    double IDF = std::log(
+    float IDF = fastapprox::fasterlog(
         1.0 + (sd.num_docs - sd.doc_count + 0.5) / (sd.doc_count + 0.5));
```
skystrife: Will this cause an up-conversion to double before down-converting to float for the argument? We may need to say things like `1.0f` and `0.5f` instead.

skystrife:
How accurate is `fasterlog`? Based on [a blog post by the author](http://www.machinedlearnings.com/2011/06/fast-approximate-logarithm-exponential.html), it seems like it's significantly worse than `fastlog`. If it's not leading to significant ranking improvements, maybe just sticking with `fastlog` is better?

src/index/ranker/okapi_bm25.cpp
smassung: Good catch, I can fix those.

smassung:
I can test it, but I'm fine with just switching to `fastlog` instead. I think both are way ahead of `std::log` in any event.

PR #85, opened 2015-05-02
Allow building in Windows with MSYS2/MinGW64

skystrife:
There is some minor nastiness in the CMakeLists that I wasn't quite sure how to handle, but everything seems to be working.


skystrife:
(Ignore AppVeyor, I've removed it for now.)

include/io/mman-win32/mman.h
```diff
@@ -0,0 +1,55 @@
+/*
```
smassung: Could you add where these files are from? Perhaps http://mman-win32.googlecode.com/svn/trunk/?

src/classify/classifier/svm_wrapper.cpp
```diff
@@ -96,9 +114,16 @@ void svm_wrapper::train(const std::vector<doc_id>& docs)
         out << idx_->liblinear_data(d_id) << "\n";
     out.close();
 
+#ifndef _WIN32
     std::string command = svm_path_ + executable_ + "train "
                           + options_.at(kernel_) + " svm-train";
     command += " > /dev/null 2>&1";
+#else
+    // see comment in classify()
+    auto command = "\"\"" + svm_path_ + executable_ + "train.exe\" "
```
smassung: Where are these .exe files coming from? Are they generated by deps/libsvm-modules somehow?

skystrife: They are generated by the ExternalProject calls in the CMakeLists.txt in the classify dir, just like they are on Linux. All executables have ".exe" added automatically (including ours), since Windows doesn't run anything without that extension.

include/io/mman-win32/mman.h
skystrife: I'm actually not sure what to do here. Google code is shutting down, so that will be a dead link soon (year or so timeline). We could maintain a mirror under the meta-toolkit organization and credit the original source from there. It doesn't seem like the original author migrated anything as far as I can tell...

src/classify/classifier/svm_wrapper.cpp
smassung: Oh ok, I see.

include/io/mman-win32/mman.h
smassung: That's right, I forgot about Google code going away... I guess we will just leave it.

PR #87, opened 2015-06-01
Change the type of query term weights to double

husseinhazimeh:
This allows updating the query using a feedback mechanism.

### DIFF
--- a/include/index/score_data.h
+++ b/include/index/score_data.h
@@ -56,7 +56,7 @@ struct score_data
     /// doc term id
     term_id t_id;
     /// query term count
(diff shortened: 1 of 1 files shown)


skystrife:
Could you also rename the field to `query_term_weight`? I think that makes it more clear that it's not actually a count but a floating point number.

husseinhazimeh:
Yes, I also think that "weights" will make the purpose behind this field clearer. The name has to be changed in all the modules that depend on score_data. I will search the project and replace all instances of "query_term_count" with "query_term_weight".

PR #105, opened 2015-09-04
Direct forward index creation

skystrife:
This PR allows us to create `forward_index` instances directly, rather than only being able to create them from `inverted_index` or from libsvm formatted data.

While seemingly minor, this change will allow us to have classification-specific analyzers that can create real-valued features, which was not possible before without using pre-generated libsvm formatted data.

This change also includes a lot of refactoring around `postings_*` and the multiway merge sort in a way that should make the code a bit prettier, and also signify intents a little better. For example, `inverted_index::postings_data_type` now explicitly states that the feature values are `uint64_t`, where `forward_index::postings_data_type` now explicitly states that they are `doubles`.

This changes the postings format for `forward_index` slightly (`total_counts` can now correctly be a `double` for `forward_index`, so we can get things like proper instance-scaling!), so you may need to regenerate indexes that you have on disk if you're using develop anywhere.


skystrife:
Oh I just love old compiler bugs. Working on a workaround for gcc < 4.8.3... I'll ping you when I think I've got this working.

skystrife:
OK, looks like I've fixed all of the CI issues...

@smassung, this is ready for review.

src/test/forward_index_test.cpp
```diff
@@ -128,9 +128,25 @@ int forward_index_tests()
...
     });
 
+    num_failed += testing::run_test("forward-index-build-uninvert", [&]()
+    {
+        system("rm -rf ceeaus-*");
+
+        // hack to inject "uninvert = true" at the top of the config file
+        auto cfg_contents = filesystem::file_text("test-config.toml");
+        cfg_contents = "uninvert = true\n" + cfg_contents;
+        filesystem::delete_file("test-config.toml");
```
smassung: So is it possible to create and set up toml objects programmatically?

smassung: Also, are the unit tests actually testing double values? I think the datasets we use are ceeaus and breast-cancer; breast-cancer has "2.0000" or "4.0000" as class labels, but no values like "2.3546". Or does it not even make sense to test this until we have regression implemented?

skystrife: Re programmatic toml: yes, for this case it is. There's a PR upstream for making this better/easier. Not sure why I did this... I'll fix.

skystrife: Re unit tests: the class labels are still strings, always. For regression, we'd have a different metadata field that would be a double pulled from the metadata table.

The double values I'm referring to are the actual values stored as the counts data in the postings file, as well as the "total counts" value stored in the postings file (which, for `inverted_index` gives you the total term occurrence count in the entire corpus, and for `forward_index` now gives you the sum of all of the weights for all of the terms in that document, which we can use for e.g. normalization).

smassung: Ok, cool. I can also change my hack in search-app to use this feature as well then.

smassung:
Travis failed on OS X because of the winnow accuracy being too low. Just reran it. I wonder if winnow is working correctly...

skystrife:
This is probably spurious. Winnow didn't change, and I know that the forward index itself is fine (if I were to switch back to using `uint64_t` for `total_counts` instead of `FeatureValue`, it produces a bit-exact `postings.index` to the tip of the develop branch).

smassung:
Oh, I'm not suggesting it was your changes that did anything. Winnow has been like this for a while (we check if the accuracy on binary classification is >65%...).

skystrife:
Is it numerically unstable for non-separable input, maybe?

smassung:
Yeah, maybe. Although the dataset is ceeaus which is clearly separable.

In any event, I'm satisfied. Merging.

PR #107, opened 2015-09-05
Merge language-model branch

smassung:
The `language-model` branch has become a `sean's-features-branch`, and I think it's in an acceptable state to merge before it gets too far off from `develop`. It would also be great to include features from `language-model` in MeTA 2.0 when `develop` itself is finally merged.

One group of additions are the `language_model`, `sentence`,  `diff`, and `diff_analyzer` classes. These all complement one another. While perfectly useable on their own, there are two things we should look into as (I suggest) separate features:

- potentially unify `sentence` and `sequence` somehow; there is a decent amount of overlap, though I can't decide whether they should be separate classes or not
- language model estimation; currently the LM requires an `.arpa` file (at least before it creates its own binary files)
- an efficient language model `top_k` function; this is used to predict the top *k* most likely tokens given a context. The way the current LM is built is optimized for ngram query speed. While possible to run the `top_k` function, it requires iterating over the unigram vocabulary. In order to solve this, some additional (most likely tree-based) data structures will probably have to be added

The other group of additions are the `feature_selector`s. These are based on the old feature selectors, except they operate on `disk_vector`s and tokenized/analyzed documents. A demo app called `feature-summary` shows one way how they can be used. Again, this is perfectly useable as-is, but would be better with:

- creating `selected_index`, or whatever you want to call it. It's another wrapped index object that uses the feature selectors' `selected(term_id)` function to check whether to search for (and cache) terms. Theoretically this should be simple enough to write, but I just haven't gotten around to it yet.

In summary, there are many new features, with four additional "wanted" features. I'd like to get the current code into develop and work on the wanted features as separate issues. @skystrife , let's see what you think.


src/sequence/crf/tools/pos_tokenizer.cpp
```diff
@@ -0,0 +1,99 @@
...
+        tagger.tag(seq);
+        for (auto& obs : seq)
+        {
+            auto word = obs.symbol();
+            if (ptb_special.find(word) != ptb_special.end())
+                std::cout << word << " ";
+            else if (keep_list.find(word) != keep_list.end())
+            {
+                word = utf::foldcase(word);
+                std::cout << word << " ";
```
smassung: `std::cout << utf::foldcase(word) << " ";` ...

include/analyzers/diff_analyzer.h
```diff
@@ -0,0 +1,63 @@
...
+    virtual void tokenize(corpus::document& doc) override;
+
+    /// Identifier for this analyzer.
+    const static std::string id;
+
+  private:
+    /// The token stream to be used for extracting tokens
+    std::unique_ptr<token_stream> stream_;
+
+    std::shared_ptr<lm::diff> diff_;
```
skystrife: Since this depends on the language model part of the library (which not everyone will necessarily want to use), this analyzer's code should probably be in the lm folder like e.g. `pos_analyzer` (sequence library) and `tree_analyzer` (parser library).

smassung: Good call, I can fix that.

include/features/selector_factory.h
```diff
@@ -0,0 +1,105 @@
...
+ * register any new selectors they write.
+ */
+template <class Selector>
+void register_selector()
+{
+    selector_factory::get().add(Selector::id,
+                                [](const cpptoml::table& config,
+                                   std::shared_ptr<index::forward_index> idx)
+                                {
+        return make_selector<feature_selector>(config, std::move(idx));
```
skystrife: Did you mean

```cpp
return make_selector<Selector>(...);
```

since `feature_selector` is abstract?

Alternatively, it's probably equivalent to saying

```cpp
selector_factory::get().add(Selector.id, make_selector<Selector>);
```

since the lambda isn't doing anything extra.

include/lm/diff.h
```diff
@@ -0,0 +1,166 @@
...
+    diff(const cpptoml::table& config);
+
+    /**
+     * Default move constructor.
+     */
+    diff(diff&&) = default;
+
+    /**
+     * @param sent The sentence object to inspect
+     * @return the index of the least-likely ngram according
```
skystrife: according to _____?

Maybe worth mentioning expected runtime?

include/lm/diff.h
```diff
@@ -0,0 +1,166 @@
...
+        return lm_;
+    }
+
+    /**
+     * @param sent The sentence to transform
+     * @param use_lm
+     * @return a sorted list of candidate corrections and their scores
+     */
+    std::vector<std::pair<sentence, double>> candidates(const sentence& sent,
+                                                        bool use_lm = true);
```
skystrife: Might be nice to document complexity here too.

include/lm/diff.h
```diff
@@ -0,0 +1,166 @@
...
+namespace meta
+{
+namespace lm
+{
+class diff
+{
+  public:
+    /**
+     * @param config_file The file containing configuration information
+     */
```
skystrife: What is the configuration information? Maybe an example toml snippet would be useful.

include/lm/diff.h
```diff
@@ -0,0 +1,166 @@
...
+
+    /// How many candidate sentences to store.
+    static constexpr uint64_t max_cand_size_ = 20;
+
+    /// Balances perplexity and edit weights.
+    static constexpr double lambda_ = 0.5;
+
+    /// Whether to insert likely words based on the language model. This is
+    /// currently turned off due to the LM representation making it inefficient.
+    static constexpr bool lm_generate_ = false;
```
skystrife: Should these three things be configurable?

include/lm/language_model.h
```diff
@@ -26,87 +51,74 @@ class language_model
...
+    float log_prob(sentence tokens) const;
 
-  private:
+    /**
+     * @param prev Seen tokens to base the next token off of
+     * @param k Number of results to return
+     * @return a sorted vector of likely next tokens
+     */
+    std::vector<std::pair<std::string, float>> top_k(const sentence& prev,
+                                                     size_t k) const;
```
skystrife: Again it might be nice to note complexity guarantees here (since I assume this is inefficient right now?).

include/lm/language_model.h
```diff
@@ -26,87 +51,74 @@ class language_model
...
 
-    /// The interpolation coefficient for smoothing LM probabilities
-    constexpr static double lambda_ = 0.7;
+    std::string prefix_;
+};
+
+class language_model_exception : public std::runtime_error
+{
+    using std::runtime_error::runtime_error;
 };
```
skystrife: I don't know what our stance is on this in general, but I prefer to have things like `language_model::exception` if possible (but that gets cumbersome with template classes).

Whatever we decide, we should probably make this consistent across the toolkit, since I don't think it currently is.

include/lm/lm_node.h
```diff
@@ -0,0 +1,52 @@
...
+
+namespace meta
+{
+namespace lm
+{
+/**
+ * Simple struct to keep track of probabilities and backoff values that is
+ * packed into a uint64_t for storage.
+ */
+struct lm_node
```
skystrife: Is this class used externally anywhere, or is it an implementation detail?

If it's an implementation detail, maybe it's good to just tuck it away in the `.cpp` file where it's needed? Then I won't feel bad about it having no comments.

include/lm/sentence.h
```diff
@@ -0,0 +1,139 @@
...
+     */
+    std::string to_string() const;
+
+    /**
+     * @param idx
+     * @return the token at the specified index
+     */
+    const std::string& operator[](size_type idx) const;
+
+    sentence operator()(size_type from, size_type to) const;
```
skystrife: Comment? I assume this is like a substring operator?

include/lm/sentence.h
```diff
@@ -0,0 +1,139 @@
...
+
+    /**
+     * @param idx
+     * @param token
+     * @param weight The weight that this edit carries
+     * @return replace the token at the specified index with the provided token
```
skystrife: "Returns a voidean value..."

include/lm/sentence.h
```diff
@@ -0,0 +1,139 @@
...
+#include <stdexcept>
+#include <deque>
+#include <vector>
+#include <string>
+
+namespace meta
+{
+namespace lm
+{
+class sentence
```
skystrife: Class comment?

include/lm/sentence.h
```diff
@@ -0,0 +1,139 @@
...
+
+    /**
+     * @return the average weight of edits to this sentence
+     */
+    double average_weight() const;
+
+    /**
+     * @return the sequence of edit weights to this sentence
+     */
+    std::vector<double> weights() const;
```
skystrife: This doesn't mean anything to me. Maybe it would if I read the paper?

Otherwise, maybe this should be elaborated a bit. Or maybe the class comment would explain. idk

include/lm/sentence.h
```diff
@@ -0,0 +1,139 @@
...
+
+    /**
+     * @return the operations (edits) performed on this sentence
+     */
+    const std::vector<std::string>& operations() const;
```
skystrife: Edits are strings? Thought they were things like insert, remove, replace, etc.

include/lm/sentence.h
```diff
@@ -0,0 +1,139 @@
...
+
+    std::string front() const;
+
+    std::string back() const;
```
skystrife: Do these two need to copy? Method comments?

include/lm/sentence.h
```diff
@@ -0,0 +1,139 @@
...
+
+    void push_front(const std::string& token);
+
+    void pop_front();
+
+    void push_back(const std::string& token);
+
+    void pop_back();
```
skystrife: Probably want to also have either overloads for `std::string&&` or `template <class... Args> void emplace_X(Args&&... args)` (with the second one being more preferable).

include/lm/static_probe_map.h
```diff
@@ -0,0 +1,93 @@
...
+
+  public:
+    /**
+     * Basic exception for static_probe_map interactions.
+     */
+    class static_probe_map_exception : public std::runtime_error
+    {
+      public:
+        using std::runtime_error::runtime_error;
+    };
```
skystrife: If you're gonna nest it, it should just be `exception`.

```cpp
catch (meta::lm::static_probe_map::static_probe_map_exception& ex)
```
is a bit of a mouthful.

src/analyzers/analyzer.cpp
```diff
@@ -6,12 +6,7 @@
 #include "analyzers/filter_factory.h"
 #include "analyzers/multi_analyzer.h"
 #include "analyzers/token_stream.h"
-#include "analyzers/filters/alpha_filter.h"
-#include "analyzers/filters/empty_sentence_filter.h"
-#include "analyzers/filters/length_filter.h"
-#include "analyzers/filters/list_filter.h"
-#include "analyzers/filters/lowercase_filter.h"
-#include "analyzers/filters/porter2_stemmer.h"
+#include "analyzers/filters/all.h"
```
skystrife: Not using `all.h` here is a compilation time optimization (don't need to parse as many headers for this translation unit since we only use a few filters in it). Any reason to switch this?

src/analyzers/diff_analyzer.cpp
```diff
@@ -0,0 +1,85 @@
...
+            else
+            {
+                for (auto& e : edits)
+                    doc.increment(e, 1);
+            }
+        }
+        catch (lm::sentence_exception& ex)
+        {
+            doc.increment("error", 1);
+        }
```
skystrife: Is this a reasonable failure handling strategy, or should things just crash if this happens? "threw an exception" as a feature seems a little weird.

src/features/feature_selector.cpp
```diff
@@ -0,0 +1,241 @@
...
+{
+namespace features
+{
+feature_selector::feature_selector(const std::string& prefix,
+                                   std::shared_ptr<index::forward_index> idx)
+    : prefix_{prefix},
+      idx_{std::move(idx)},
+      selected_{prefix_ + ".selected", idx_->unique_terms()}
+{ /* nothing */
+}
```
skystrife: This bugs me, can it be

```cpp
{
    /* nothing */
}
```

or

```cpp
{
    // nothing
}
```

(yes, this is very pedantic)

include/features/feature_selector.h
```diff
@@ -0,0 +1,197 @@
...
+    feature_selector(const std::string& prefix,
+                     std::shared_ptr<index::forward_index> idx);
+
+    /**
+     * Creates the state of this feature_selector if necessary; this logic is
+     * outside the constructor since it requires pure virtual functions
+     * implemented by deriving classes.
+     * @param features_per_class
+     */
+    void init(uint64_t features_per_class);
```
skystrife: Do users have to remember to call this, or does it happen automatically?

src/features/feature_selector.cpp
```diff
@@ -0,0 +1,241 @@
...
+        for (uint64_t lbl = 0; lbl < idx_->num_labels(); ++lbl)
+            scores[lbl][tid] = std::make_pair(
+                tid, score(static_cast<label_id>(lbl + 1), term_id{tid}));
+    }
+    prog.end();
+
+    parallel::parallel_for(
+        scores.begin(), scores.end(), [&](std::vector<pair_t>& v)
+        {
+            std::sort(v.begin(), v.end(), [&](const pair_t& a, const pair_t& b)
```
skystrife: The capture group can be empty here since it doesn't refer to anything.

I don't think it matters, but it could?

src/features/feature_selector.cpp
```diff
@@ -0,0 +1,241 @@
...
+        });
+
+    for (uint64_t lbl = 0; lbl < idx_->num_labels(); ++lbl)
+    {
+        // write (term_id, score) pairs
+        std::ofstream out{prefix_ + "." + std::to_string(lbl + 1)};
+        for (auto& score : scores[lbl])
+        {
+            io::write_binary(out, score.first);
+            io::write_binary(out, score.second);
```
skystrife: This can be really fragile. Can you use `io::packed::write()` instead?

src/features/feature_selector.cpp
```diff
@@ -0,0 +1,241 @@
...
+            std::sort(v.begin(), v.end(), [&](const pair_t& a, const pair_t& b)
+                      {
+                          return a.second > b.second;
+                      });
+        });
+
+    for (uint64_t lbl = 0; lbl < idx_->num_labels(); ++lbl)
+    {
+        // write (term_id, score) pairs
+        std::ofstream out{prefix_ + "." + std::to_string(lbl + 1)};
```
skystrife: add `std::ios::binary` as a second argument if you have any ofstream you're writing binary to.

...because Windows.

src/features/feature_selector.cpp
```diff
@@ -0,0 +1,241 @@
...
+}
+
+void feature_selector::select(uint64_t features_per_class /* = 20 */)
+{
+    term_id id;
+    double score;
+    std::unordered_set<term_id> terms;
+    for (uint64_t lbl = 0; lbl < idx_->num_labels(); ++lbl)
+    {
+        std::ifstream in{prefix_ + "." + std::to_string(lbl + 1)};
```
skystrife: `std::ios::binary`

src/features/feature_selector.cpp
```diff
@@ -0,0 +1,241 @@
...
+        }
+    }
+
+    // zero out old vector
+    for (auto& b : selected_)
+        b = false;
+
+    // select new features
+    for (auto& term : terms)
+        selected_[term] = true;
```
skystrife: If this is all `terms` is used for, is there any reason to make the temporary at all? Zero as the first step, then change line 95 to `selected_[id] = true;`?

src/features/feature_selector.cpp
```diff
@@ -0,0 +1,241 @@
...
+    term_id id;
+    double score;
+    std::unordered_set<term_id> terms;
+    for (uint64_t lbl = 0; lbl < idx_->num_labels(); ++lbl)
+    {
+        std::ifstream in{prefix_ + "." + std::to_string(lbl + 1)};
+        for (uint64_t i = 0; i < features_per_class; ++i)
+        {
+            io::read_binary(in, id);
+            io::read_binary(in, score);
```
skystrife: Same thing as for `write_binary`, these are really fragile. Can `packed::read` be used?

src/features/feature_selector.cpp
```diff
@@ -0,0 +1,241 @@
...
+
+bool feature_selector::selected(term_id term) const
+{
+    return selected_[term];
+}
+
+void feature_selector::select_percent(double p /* = 0.05 */)
+{
+    double num_features = p * idx_->unique_terms();
+    uint64_t per_class = num_features / idx_->num_labels(); // truncate to int
```
skystrife: Should it be an error if no features will be selected?

src/features/feature_selector.cpp
```diff
@@ -0,0 +1,241 @@
...
+    select(per_class);
+}
+
+void feature_selector::calc_probs()
+{
+    printing::progress prog{" > Calculating feature probs: ", idx_->num_docs()};
+    uint64_t total_terms = 0;
+    for (doc_id did = doc_id{0}; did < idx_->num_docs(); ++did)
```
skystrife:
```cpp
for (doc_id did{0}; did < ....)
```

src/features/feature_selector.cpp
```diff
@@ -0,0 +1,241 @@
...
+
+    for (auto& p : class_prob_)
+        p /= idx_->num_docs();
+
+    for (auto& p : term_prob_)
+        p /= total_terms;
+
+    for (auto& probs : co_occur_)
+        for (auto& p : probs)
+            p /= total_terms;
```
skystrife: (pedantic, but `probs` can be `const auto&` here)

src/features/feature_selector.cpp
```diff
@@ -0,0 +1,241 @@
...
+                  << idx_->class_label_from_id(static_cast<label_id>(lbl + 1))
+                  << "\":" << std::endl
+                  << "===============================" << std::endl;
+
+        // read (term_id, score) pairs
+        std::ifstream in{prefix_ + "." + std::to_string(lbl + 1)};
+        for (uint64_t i = 0; i < k; ++i)
+        {
+            io::read_binary(in, tid);
+            io::read_binary(in, score);
```
skystrife: As usual, `read/write_binary` are scary. `packed`?

src/features/feature_selector.cpp
```diff
@@ -0,0 +1,241 @@
...
+    for (uint64_t lbl = 0; lbl < idx_->num_labels(); ++lbl)
+    {
+        std::cout << std::endl
+                  << "Top " << k << " features for \""
+                  << idx_->class_label_from_id(static_cast<label_id>(lbl + 1))
+                  << "\":" << std::endl
+                  << "===============================" << std::endl;
+
+        // read (term_id, score) pairs
+        std::ifstream in{prefix_ + "." + std::to_string(lbl + 1)};
```
skystrife: `std::ios::binary`

src/features/feature_selector.cpp
```diff
@@ -0,0 +1,241 @@
...
+    for (auto& probs : co_occur_)
+        for (auto& p : probs)
+            p /= total_terms;
+}
+
+void feature_selector::print_summary(uint64_t k /* = 20 */) const
+{
+    term_id tid;
+    double score;
+    for (uint64_t lbl = 0; lbl < idx_->num_labels(); ++lbl)
```
skystrife: I've seen this a few times, is it more clean to just say

```cpp
for (auto lbl = 1_lid, lbl <= idx_->num_labels(); ++lbl)
```

?

src/features/feature_selector.cpp
```diff
@@ -0,0 +1,241 @@
...
+    {
+        std::cout << std::endl
+                  << "Top " << k << " features for \""
+                  << idx_->class_label_from_id(static_cast<label_id>(lbl + 1))
```
skystrife: ...because then you don't need the +1 or the `static_cast`.

src/features/information_gain.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+    // if any denominators were zero, make the expression zero
+    if (std::isnan(gain_tc))
+        gain_tc = 0.0;
+    if (std::isnan(gain_ntnc))
+        gain_ntnc = 0.0;
+    if (std::isnan(gain_ntc))
+        gain_ntc = 0.0;
+    if (std::isnan(gain_tnc))
+        gain_tnc = 0.0;
```
skystrife: Is checking `isnan` good enough here? We're not going to deal with weirdness like +/- infinity or anything, right?

src/features/information_gain.cpp
```diff
@@ -0,0 +1,44 @@
...
+    double p_tnc = term_and_not_class(tid, lid);
+    double p_c = prob_class(lid);
+    double p_t = prob_term(tid);
+    double p_nc = 1.0 - p_c;
+    double p_nt = 1.0 - p_t;
+
+    double gain_tc = p_tc * std::log(p_tc / (p_t * p_c));
+    double gain_ntnc = p_ntnc * std::log(p_ntnc / (p_nt * p_nc));
+    double gain_ntc = p_ntc * std::log(p_ntc / (p_nt * p_c));
+    double gain_tnc = p_tnc * std::log(p_tnc / (p_t * p_nc));
```
skystrife: Is it worthwhile to use `fastapprox` here?

src/features/odds_ratio.cpp
```diff
@@ -0,0 +1,28 @@
...
+    double p_tc = term_and_class(tid, lid);
+    double p_tnc = term_and_not_class(tid, lid);
+    double numerator = p_tc * (1.0 - p_tnc);
+    double denominator = (1.0 - p_tc) * p_tnc;
+
+    // avoid divide by zero
+    if (denominator == 0.0)
+        return 0.0;
+
+    return std::log(numerator / denominator);
```
skystrife: Again, candidate for `fastapprox` maybe?

src/features/selector_factory.cpp
```diff
@@ -0,0 +1,59 @@
...
+    auto prefix = table->get_as<std::string>("prefix");
+    if (!prefix)
+        throw selector_factory::exception{"no prefix in [features] table"};
+
+    auto method = table->get_as<std::string>("method");
+    if (!method)
+        throw selector_factory::exception{
+            "feature selection method required in [features] table"};
+
+    uint64_t features_per_class = 20;
```
skystrife: Is there some constexpr somewhere you could read this from instead, so if this default changes we don't forget to update it here too?

src/lm/diff.cpp
```diff
@@ -0,0 +1,255 @@
...
+        throw diff_exception{"n-value not specified in config"};
+    n_val_ = *nval;
+
+    auto edits = table->get_as<int64_t>("max-edits");
+    if (!edits)
+        throw diff_exception{"max-edits not specified in config"};
+    max_edits_ = *edits;
+
+    auto b_pen = table->get_as<double>("base-penalty");
+    base_penalty_ = b_pen ? *b_pen : 0.0;
```
skystrife: I should probably add `value_or(T)` to `cpptoml::option<T>` (and `meta::util::optional<T>`)...

src/lm/diff.cpp
```diff
@@ -0,0 +1,255 @@
...
+    diff::candidates(const sentence& sent, bool use_lm /* = false */)
+{
+    use_lm_ = use_lm;
+    using pair_t = std::pair<sentence, double>;
+    auto comp = [](const pair_t& a, const pair_t& b)
+    {
+        return a.second < b.second;
+    };
+    std::priority_queue<pair_t, std::vector<pair_t>, decltype(comp)> candidates{
+        comp};
```
skystrife: oh God what

I think this needs to be its own type.

skystrife: Wait, is *this* the reason for all of the `template <class PQ>` later?

src/lm/diff.cpp
```diff
@@ -0,0 +1,255 @@
...
+    add(candidates, sent);
+
+    seen_.clear();
+    step(sent, candidates, 0);
+
+    std::vector<pair_t> sorted;
```
skystrife: `sorted.reserve(candidates.size())`?

src/lm/diff.cpp
```diff
@@ -0,0 +1,255 @@
...
+
+template <class PQ>
+void diff::add(PQ& candidates, const sentence& sent)
+{
+    seen_.insert(sent.to_string());
+    auto score = lambda_ * lm_.perplexity_per_word(sent)
+                 + (1.0 - lambda_) * sent.average_weight();
+    candidates.emplace(sent, score);
+    if (candidates.size() > max_cand_size_)
+        candidates.pop();
```
skystrife: We have this pattern a lot. Is there a `util::fixed_heap<T, Compare>` class hiding?

src/lm/diff.cpp
```diff
@@ -0,0 +1,255 @@
...
+        auto ngram = sent(i - n_val_, i);
+        auto prob = lm_.log_prob(ngram);
+        if (prob < min_prob)
+        {
+            min_prob = prob;
+            best_idx = i - 1;
+        }
+    }
+
+    return best_idx;
```
skystrife: Wait, so what is this returning? The index of the word that ends the least likely ngram?

src/lm/diff.cpp
```diff
@@ -0,0 +1,255 @@
...
+                sentence sub_cpy{sent};
+                sub_cpy.substitute(best_idx, next.first,
+                                   base_penalty_ + substitute_penalty_);
+
+                if (seen_.find(sub_cpy.to_string()) == seen_.end())
+                {
+                    add(candidates, sub_cpy);
+                    step(sub_cpy, candidates, depth + 1);
+                }
+            }
```
skystrife: Curious: have you profiled this part? Wondering if all these sentence copies are expensive. Is it worth having a more optimized `sentence` data structure that lets you model these edits better without the copy?

src/lm/diff.cpp
```diff
@@ -0,0 +1,255 @@
...
+            // ignore if there are no transitions found
+        }
+    }
+}
+
+template <class PQ>
+void diff::insert(const sentence& sent, size_t idx, PQ& candidates,
+                  uint64_t depth)
+{
+    for (auto& fw : fwords_)
```
skystrife: Is `fw` changed by this loop? If not, it should be `const auto&` to reflect that.

src/lm/diff.cpp
```diff
@@ -0,0 +1,255 @@
...
+void diff::set_stems(const cpptoml::table& config)
+{
+    std::unordered_set<std::string> vocab;
+    auto prefix = *config.get_as<std::string>("prefix");
+    auto dataset = *config.get_as<std::string>("dataset");
+    std::ifstream in{prefix + "/" + dataset + "/" + dataset + ".dat"};
+    std::string token;
+    while (in >> token)
+    {
+        std::transform(token.begin(), token.end(), token.begin(), ::tolower);
```
skystrife: `utf::foldcase`, in general.

src/lm/diff.cpp
```diff
@@ -0,0 +1,255 @@
...
+        vocab.insert(token);
+    }
+
+    for (auto& t : vocab)
```
skystrife: `const auto&` if `t` is not changed.

src/lm/language_model.cpp
```diff
@@ -7,213 +7,225 @@
...
+    {
+        return a.second > b.second;
+    };
+    std::vector<pair_t> candidates;
 
-    double cur = 0.0;
-    for (auto& end : it->second)
+    sentence candidate = prev;
+    candidate.push_back("word"); // the last item is replaced each iteration
+    for (auto& word : vocabulary_)
```
skystrife: `const auto&` if `word` is immutable.

src/lm/language_model.cpp
```diff
@@ -7,213 +7,225 @@
...
-        if (cur > random)
-            return end.first;
+        auto candidate = sentence{prev.to_string() + " " + word};
+        candidates.emplace_back(word, log_prob(candidate));
+        std::push_heap(candidates.begin(), candidates.end(), comp);
+        if (candidates.size() > k)
+        {
+            std::pop_heap(candidates.begin(), candidates.end(), comp);
+            candidates.pop_back();
+        }
```
skystrife: Another candidate for `util::fixed_heap`?

src/lm/language_model.cpp
```diff
@@ -7,213 +7,225 @@
...
-    std::deque<std::string> ngram;
-    for (size_t n = 1; n < N_; ++n)
-        ngram.push_back("<s>");
-
-    // keep generating until we see </s>
-    std::string output;
-    std::string next = next_token(ngram, rdist(gen));
-    while (next != "</s>")
+    std::string word;
+    std::ifstream unigrams{prefix_ + "0.strings"};
```
skystrife: Is there any particular reason this is stored uncompressed?

src/lm/sentence.cpp
```diff
@@ -0,0 +1,177 @@
...
+    if (tokenize)
+    {
+        using namespace analyzers;
+        std::unique_ptr<token_stream> stream;
+        stream = make_unique<tokenizers::icu_tokenizer>();
+        stream = make_unique<filters::empty_sentence_filter>(std::move(stream));
+        std::string text_copy{text}; // consider changing parameter to non-const
+        stream->set_content(std::move(text_copy));
+        while (*stream)
+            tokens_.push_back(stream->next());
```
skystrife: I think you'd be better off just using `utf::segmenter` directly here instead of going through the analyzer pipeline interface. This will eliminate the unnecessary copy of the `std::string` parameter, since `utf::segmenter` just takes a `util::string_view`, and I think the code will be simpler.

src/lm/sentence.cpp
```diff
@@ -0,0 +1,177 @@
...
+std::string sentence::to_string() const
+{
+    std::string result{""};
+    if (tokens_.empty())
+        return result;
+
+    for (auto& token : tokens_)
+        result += token + " ";
+
+    return result.substr(0, result.size() - 1); // remove trailing space
```
skystrife: All this string copying is probably really bad for performance. I think this class is a perfect optimization point for your `lm::diff` stuff. Here's what I'm thinking:

Instead of having your hash tables/sets having `std::string` as their keys (which you're comparing against some `temp_sentence.to_string()`), just use `sentence` objects directly. Compute the hash code of a `sentence` via a hash combine of all of its individual tokens' hash value (easy with `util::murmur_hash<>`), and equality comparison is just "do these have the same tokens in the same places". Then, when you've got a candidate `sentence` you want to check that you haven't seen yet, you can just look it up in your table directly and forgo the conversion to `std::string`.

src/lm/sentence.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+std::string sentence::front() const
+{
+    return tokens_.front();
+}
+
+std::string sentence::back() const
+{
+    return tokens_.back();
+}
```
skystrife: Yeah, these two should be either `util::string_view` or `const std::string&`.

src/lm/static_probe_map.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+namespace meta
+{
+namespace lm
+{
+static_probe_map::static_probe_map(const std::string& filename,
+                                   uint64_t num_elems)
+    : table_{filename, static_cast<uint64_t>((num_elems / 0.7) * 2)},
+      hash_{seed_}
+// load factor of 0.7; x2 for keys and vals
```
skystrife: Maybe `ceil(n / 0.7)` instead of `floor(n / 0.7)`?

include/lm/lm_node.h
```diff
@@ -0,0 +1,52 @@
...
+    lm_node(float p, float b) : prob{p}, backoff{b}
+    {
+    }
+
+    lm_node(uint64_t packed)
+    {
+        uint32_t buf = packed >> 32;
+        std::memcpy(&prob, &packed, sizeof(float));
+        std::memcpy(&backoff, &buf, sizeof(float));
+    }
```
skystrife: I don't think this works, see below where you actually write it.

src/lm/static_probe_map.cpp
```diff
@@ -0,0 +1,70 @@
...
+    {
+        if (table_[idx] == uint64_t{0})
+        {
+            table_[idx] = hashed;
+
+            // pack prob and float into uint64_t slot next to key val
+            uint64_t buf = 0;
+            std::memcpy(&table_[idx + 1], &prob, sizeof(float));
+            std::memcpy(&buf, &backoff, sizeof(float));
+            table_[idx + 1] |= (buf << 32);
```
skystrife: Wait, is this doing the right thing? Is this assuming a specific endianness?

Below is a program that (I think) is testing how you're doing this. It's got another function that's written how I would have done it.

```cpp
#include <iostream>
#include <vector>
#include <cstdint>
#include <cstring>

void seanbits()
{
    uint64_t bytes;
    float prob = 0.2;
    float backoff = 0.5;

    std::memcpy(&bytes, &prob, sizeof(float));
    uint64_t buf;
    std::memcpy(&buf, &backoff, sizeof(float));
    bytes |= (buf << 32);

    uint32_t outbuf = (bytes >> 32);
    float probout;
    float backoffout;
    std::memcpy(&probout, &bytes, sizeof(float));
    std::memcpy(&backoff, &outbuf, sizeof(float));

    std::cout << "prob: " << prob << ", out: " << probout << "\n"
              << "back: " << backoff << ", out: " << backoffout << std::endl;
}

void chasebits()
{
    uint64_t bytes;
    float prob = 0.2;
    float backoff = 0.5;

    char* buf = reinterpret_cast<char*>(&bytes);
    std::memcpy(buf, &prob, sizeof(float));
    std::memcpy(buf + sizeof(float), &backoff, sizeof(float));

    float probout;
    float backoffout;
    std::memcpy(&probout, buf, sizeof(float));
    std::memcpy(&backoff, buf + sizeof(float), sizeof(float));

    std::cout << "prob: " << prob << ", out: " << probout << "\n"
              << "back: " << backoff << ", out: " << backoffout << std::endl;
}

int main()
{
    seanbits();
    chasebits();
}
```

Output:
```
prob: 0.2, out: 0.2
back: 0.5, out: 0
prob: 0.2, out: 0.2
back: 0.5, out: 0.5
```

Now, I'm not for sure if the error is in reading or writing, but I think the way I'm writing it is more clear about the intent (and should(?) work regardless of endianness, provided that the model files are generated and read on the same machine).

In any case, this logic should probably move to `lm_node`, which has a `write_packed()` or something so that this logic is all in one place and can be easily unit tested.

src/lm/tools/diff_test.cpp
```diff
@@ -0,0 +1,68 @@
...
+    auto num_sentences = filesystem::num_lines(argv[2]);
+    printing::progress prog{"Editing sentences ", num_sentences};
+    std::ofstream out{std::string{argv[2]} + ".out"};
+    std::ofstream log{std::string{argv[2]} + ".log"};
+    std::string line;
+    size_t done = 0;
+    double do_nothing = 0;
+    while (in)
+    {
+        std::getline(in, line);
```
skystrife:
```cpp
while (std::getline(in, line))
```

src/test/features_test.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+namespace
+{
+template <class Index>
+void test_construction(Index& idx, const std::string& id)
+{
+    std::ofstream fconfig{"feature-config.toml"};
+    fconfig << "[features]\nmethod = \"" << id << "\"\n"
+            << "prefix = \"test-features\"";
+    fconfig.close();
```
skystrife: I'll try to make this easier on the cpptoml side.

src/test/lm_test.cpp
```diff
@@ -0,0 +1,50 @@
...
+        lm::sentence s2{
+            "<s> I disagree with this octopus for several reasons . </s>",
+            false};
+        lm::sentence s3{"<s> Hello world ! </s>", false};
+        lm::sentence s4{"<s> xyz xyz xyz </s>", false};
+
+        ASSERT_APPROX_EQUAL(model.log_prob(s1), -5.0682507);
+        ASSERT_APPROX_EQUAL(model.log_prob(s2), -11.7275571);
+        ASSERT_APPROX_EQUAL(model.log_prob(s3), -11.07649517);
+        ASSERT_APPROX_EQUAL(model.log_prob(s4), -16.41804123);
```
skystrife: Were these calculated by hand or is this just a regression test for yourself?

src/topics/tools/topic_corpus.cpp
```diff
@@ -0,0 +1,139 @@
...
+    return 1;
+}
+
+std::vector<size_t> get_topic_ids(std::ifstream& thetas)
+{
+    std::vector<size_t> topic_ids;
+    std::string line;
+    while (thetas)
+    {
+        std::getline(thetas, line);
```
skystrife:
```cpp
while (std::getline(thetas, line))
```

src/topics/tools/topic_corpus.cpp
```diff
@@ -0,0 +1,139 @@
...
+        std::istringstream stream{line};
+        size_t did;
+        stream >> did;
+
+        std::string to_split;
+        size_t best_topic = 0;
+        double best_prob = 0;
+        while (stream)
+        {
+            stream >> to_split;
```
skystrife: while (stream >> to_split)

src/topics/tools/topic_corpus.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+void create_topic_corpus(const std::string& prefix, const std::string& dataset,
+                         std::ifstream& thetas)
+{
+    auto topic_ids = get_topic_ids(thetas);
+    auto new_file = prefix + "/" + dataset + "/" + dataset
+                    + "-topics.dat.labels";
+    std::ofstream out_labels{new_file};
+    size_t num_topics = 0;
+    for (auto& topic : topic_ids)
```
skystrife: `const auto&`

src/topics/tools/topic_corpus.cpp
```diff
@@ -0,0 +1,139 @@
...
+    std::vector<std::string> orig_labels;
+    std::string buf;
+    while (orig_labels_in >> buf)
+        orig_labels.push_back(buf);
+
+    std::cout << orig_labels.size() << std::endl;
+
+    std::unordered_map<std::string, std::vector<double>> counts;
+    uint64_t idx = 0;
+    for (auto& topic : topic_ids)
```
skystrife: `const auto&`

src/topics/tools/topic_corpus.cpp
```diff
@@ -0,0 +1,139 @@
...
+    {
+        auto label = orig_labels[idx];
+        if (counts[label].empty())
+            counts[label].resize(num_topics, 0.0);
+        ++counts[label][topic];
+        ++idx;
+    }
+
+    std::ofstream out_dist{dataset + ".topic-dist"};
+    for (auto& label : counts)
```
skystrife: `const auto&`

skystrife:
Oh, and since the paper was accepted, you should add a link to it in the class comment for the diff stuff. =)

include/lm/lm_node.h
smassung: It's used by `static_probe_map`, but I can't forward-declare it since `optional` needs the entire class. I'll add some comments to the member functions and keep it separate.

src/features/feature_selector.cpp
smassung: I blame clang-format

src/features/feature_selector.cpp
smassung: Yeah, these are all changed.

src/features/feature_selector.cpp
smassung: Not sure why I did it this way. Changed to your suggestion, which makes more sense.

src/features/feature_selector.cpp
smassung: Added some error handling.

src/features/feature_selector.cpp
smassung: `p` is part of `probs`, so `probs` is being modified

src/features/feature_selector.cpp
smassung: Yep, changed

src/lm/language_model.cpp
smassung: We should probably make this class

src/lm/language_model.cpp
smassung: It's relatively small, and for debugging when I was writing it.

But.. it also allows a nice hack for `diff` if I set the function words to point to this file. Then I can insert all words in the vocab easily. (Since it's a user-specified file, it should be human-readable...)

src/lm/sentence.cpp
smassung: Good idea, I can modify this.

src/lm/static_probe_map.cpp
smassung: I think 0.7 is a high-enough alpha that it probably doesn't really matter.

src/lm/static_probe_map.cpp
smassung: I'll look into this

src/test/lm_test.cpp
smassung: It's based on output from kenlm

```
I=3 2 -1.0685297	disagree=4 3 -1.0058641	with=5 3 -0.2141374	this=6 3 -0.09438104	statement=7 3 -0.1721986	for=8 3 -0.42091483	several=9 3 -1.6239449	reasons=10 3 -0.17254661	.=11 3 -0.295592	</s>=2 3 -0.00014224081	Total: -5.0682507 OOV: 0
I=3 2 -1.0685297	disagree=4 3 -1.0058641	with=5 3 -0.2141374	this=6 3 -0.09438104	octopus=0 1 -5.0185504	for=8 1 -1.9226898	several=9 2 -1.9351243	reasons=10 3 -0.17254661	.=11 3 -0.295592	</s>=2 3 -0.00014224081	Total: -11.727557 OOV: 1
Hello=0 1 -4.7541533	world=580 1 -3.0229268	!=821 1 -3.170956	</s>=2 2 -0.12845917	Total: -11.076495 OOV: 1
xyz=0 1 -4.7541533	xyz=0 1 -4.1183567	xyz=0 1 -4.1183567	</s>=2 1 -3.4271748	Total: -16.418041 OOV: 3

include/lm/diff.h
smassung: Yes

src/analyzers/diff_analyzer.cpp
smassung: If there are no possible candidates able to be generated from the LM, the LM throws an exception. I can rename the feature to make this clear.

include/features/selector_factory.h
smassung: Fixed

include/lm/diff.h
smassung: Fixed

include/lm/diff.h
smassung: Fixed

include/lm/diff.h
smassung: Fixed

include/lm/language_model.h
smassung: Fixed

include/lm/language_model.h
smassung: I think outside is better. Starting to move my exceptions out.

include/lm/sentence.h
smassung: Yep, fixed.

include/lm/sentence.h
smassung: Yes, it's a voidean value. Fixed!

include/lm/sentence.h
smassung: Fixed

include/lm/sentence.h
smassung: Fixed

include/lm/sentence.h
smassung: Fixed

include/lm/sentence.h
smassung: Fixed, const ref.

include/lm/sentence.h
smassung: Fixed, forwarded args.

include/lm/static_probe_map.h
smassung: Moved outside class.

src/analyzers/analyzer.cpp
smassung: Nope, fixed.

include/features/feature_selector.h
smassung: Can't get it to compile for some reason with the templates, might have to ask you to look at it...

src/features/feature_selector.cpp
smassung: Fixed

src/features/feature_selector.cpp
smassung: Yeah, fixed.

src/features/feature_selector.cpp
smassung: Fixed

src/features/feature_selector.cpp
smassung: Fixed

src/features/feature_selector.cpp
smassung: Fixed

src/features/feature_selector.cpp
smassung: Fixed

src/features/feature_selector.cpp
smassung: Fixed

src/features/feature_selector.cpp
smassung: Fixed

src/features/information_gain.cpp
smassung: Changed to `std::isfinite` which should check for all those cases.

src/features/information_gain.cpp
smassung: Need to investigate, but probably!

src/lm/diff.cpp
smassung: That would be great!

src/features/selector_factory.cpp
smassung: Isn't this the only place it occurs? I might be missing something. I don't see it anywhere in `selector.h`, except in the default params, but it doesn't necessarily have to be the same.

src/lm/diff.cpp
smassung: Yeah, we need `util::fixed_heap`. I have a habit of making things templates when I don't want to write out the type, which probably means the type shouldn't exist...

src/lm/diff.cpp
smassung: Candidates should be small enough that it doesn't matter (e.g. 20), but I can add it anyway.

src/lm/diff.cpp
smassung: Agree, need to make one

src/lm/diff.cpp
smassung: Correct, updated comment in header to specify this.

src/lm/diff.cpp
smassung: Profiling research code? But no, seriously, I haven't profiled anything about this yet.

src/lm/diff.cpp
smassung: Fixed

src/lm/diff.cpp
smassung: Fixed

src/lm/diff.cpp
smassung: Fixed

src/lm/language_model.cpp
smassung: Fixed

src/lm/sentence.cpp
smassung: Fixed

src/lm/tools/diff_test.cpp
smassung: Fixed

src/test/features_test.cpp
smassung: Awesome

src/topics/tools/topic_corpus.cpp
smassung: Fixed

src/topics/tools/topic_corpus.cpp
smassung: Fixed

src/topics/tools/topic_corpus.cpp
smassung: Fixed

src/topics/tools/topic_corpus.cpp
smassung: Fixed

src/topics/tools/topic_corpus.cpp
smassung: Fixed

smassung:
Most things taken care of. To do:
- inspect whether there is a bug with the packed `uint64_t` for `lm_node` and move all logic into there
- make `util::fixed_heap`
- store hashes of `lm::sentence` where appropriate instead of calling `to_string` all the time
- use utf segmenter instead of invoking filter chains
- try to make `selector::init` a friend of `selector_factory`

Hopefully I can do these later on Monday and we can merge this branch, and then merge `typed-analyzers`.

smassung:
Ok, looks like I might have broken something in `feature_select`. Adding that to the list of things to look at.

smassung:
Fixed the features issue, working on other ones.

Issue #106, opened 2015-09-04
Support non-double feature values in analyzers


skystrife:
Currently the `document` and `analyzer` class assumes that counts are stored as `double` values always. This isn't the case, particularly for `inverted_index` classes.

The change here will be to make `analyzer` be `analyzer<FeatureValue>`, with `FeatureValue` taking values `double` (for `forward_index`) and `uint64_t`. Tokenizing a document will return a `std::unordered_map<std::string, FeatureValue>` instead of the counts being stored directly in the `document` class.

This will make it more readily obvious what kinds of analyzers are appropriate for what kinds of index classes. Adding a feature for classification that sets real-valued features is fine, but it is not appropriate for creating `inverted_index` instances. We can now allow for this and error out appropriately if someone uses an inappropriate analyzer for their chosen index type. Analyzers that output integer values feature counts could be used for either `inverted_index` or `forward_index` creation.

While `analyzer` will become a template class, since we only will have two template instantiations for most of our analyzer classes, we can keep the analyzers compiled as regular static libraries as before (rather than becoming fully header-only and reparsed for each translation unit that includes them).

skystrife:
Fixed with the merge of PR #108.

PR #108, opened 2015-09-06
Typed analyzers

skystrife:
This addresses issue #106.


CMakeLists.txt
```diff
@@ -237,6 +237,27 @@ if (META_HAS_NONEMPTY_HASH_SUPPORT)
                              -DMETA_HAS_NONEMPTY_HASH_SUPPORT)
 endif()
 
+# work around a bug in gcc 4.8.1 where having a lambda as a default
+# argument to a member function template causes an internal compiler error
+check_cxx_source_compiles("
+struct h {
+  template <class Function = void(*)()>
+  void foo(Function&& fun = [](){}) {
+    fun();
```
smassung: [](){}

include/analyzers/analyzer.h
```diff
@@ -48,71 +69,88 @@ class analyzer
...
 
-    /**
-     * @param src The token stream that will feed into this filter
-     * @param config The config group used to create the filter from
-     * @return a single filter specified by a config object
-     */
-    static std::unique_ptr<token_stream>
-        load_filter(std::unique_ptr<token_stream> src,
-                    const cpptoml::table& config);
+extern template std::unique_ptr<analyzer<uint64_t>>
```
smassung: Can you explain what the `extern`s accomplish here? (For documentation)

src/analyzers/ngram/ngram_word_analyzer.cpp
```diff
@@ -16,52 +16,61 @@ namespace meta
 namespace analyzers
 {
 
-const std::string ngram_word_analyzer::id = "ngram-word";
+template <class T>
+const std::string ngram_word_analyzer<T>::id = "ngram-word";
```
smassung: We can probably create just `ngram_analyzer` and have it use different input streams, for example unigram words or POS tags. Not for this PR, but in general.

smassung:
We're going to have a lot of documentation rewriting on meta-toolkit.org for 2.0...

include/analyzers/analyzer.h
skystrife: So these show up all over the place now. Basically every `analyzer` header file and every `tree_featurizer` header file. Do you want me to document it everywhere, or just here?

Essentially what this is doing is telling the compiler to not attempt to instantiate the template for the specified type in all translation units that include this header. This would obviously cause an error at link time unless you link with an object file that actually has the instantiation. We force the compiler to instantiate the templates *once* in the corresponding `.cpp` file. This is what saves us from having to move everything to be header-only and parse and instantiate every function that's used in every translation unit.

Hmm... I guess an easier explanation is that this:

```cpp
extern template class myanalyzer<uint64_t>;
```

is the declaration and 

```cpp
template class myanalyzer<uint64_t>;
```

is the definition, to use the standard terms for regular classes. See: http://en.cppreference.com/w/cpp/language/class_template#Explicit_instantiation

**tl;dr:** Not sure how you want me to document what this is doing, beyond something like


```cpp
// declare the valid instantiations for this analyzer
extern template class ngram_word_analyzer<uint64_t>;
extern template class ngram_word_analyzer<double>;

// declare the valid instantiations for this analyzer's trait class
extern template struct analyzer_traits<ngram_word_analyzer<uint64_t>>;
extern template struct analyzer_traits<ngram_word_analyzer<double>>;
```

smassung: I think the one-liner comments are good since this would seem a little unusual for someone that happens to see this.

smassung:
I'm going to merge this; the travis failures are spurious.

PR #112, opened 2015-09-12
language_model optimizations

smassung:
This improves the `language_model` operation running times. As a test on my own computer, these changes made `diff-test` run in 1m12s compared to 1m55s originally on the CoNLL data.

### DIFF
--- a/include/lm/language_model.h
+++ b/include/lm/language_model.h
@@ -23,6 +23,11 @@ namespace meta
 {
 namespace lm
 {
+class language_model_exception : public std::runtime_error
+{
+    using std::runtime_error::runtime_error;
+};
+
 /**
  * A very simple language model class that reads existing language model data
  * from a .arpa file. Currently, estimation of the language model is not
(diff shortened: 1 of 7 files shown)


include/lm/token_list.h
```diff
@@ -0,0 +1,118 @@
...
+}
+
+template <class HashAlgorithm>
+void hash_append(HashAlgorithm& h, const token_list& list)
+{
+    using util::hash_append;
+    for (const auto& val : list.tokens())
+        hash_append(h, val);
+    hash_append(h, list.size());
+}
```
skystrife: Potentially faster (probably doesn't matter in practice):

```cpp
template <class HashAlgorithm>
void hash_append(HashAlgorithm& h, const token_list& list)
{
    h(tokens_.data(), sizeof(term_id) * list.size());
    hash_append(h, list.size());
}
```

(You'll need to friend your `hash_append` overload to do this, but that's easy.)

include/lm/language_model.h
```diff
@@ -111,9 +117,11 @@ class language_model
 
     std::vector<static_probe_map> lm_;
 
-    std::vector<std::string> vocabulary_;
+    std::unordered_map<std::string, term_id> vocabulary_;
```
skystrife: `util::probe_set` is another option.

src/lm/static_probe_map.cpp
```diff
@@ -56,10 +56,10 @@ util::optional<lm_node> static_probe_map::find(const std::string& key) const
     }
 }
 
-uint64_t static_probe_map::hash(const std::string& str) const
+uint64_t static_probe_map::hash(const token_list& tokens) const
 {
     util::murmur_hash<> hasher{seed_};
-    hasher(str.data(), str.length());
+    hasher(tokens.tokens().data(), tokens.size() * sizeof(term_id));
```
skystrife: aka `util::hash_append(hasher, tokens)`

include/lm/token_list.h
smassung: Ok, changed to that.

smassung:
Wow, changing the hash stuff actually made `diff-test` take 1m0s as opposed to 1m12s.

smassung:
Nevermind, that was using the old lm files that weren't regenerated. It's about the same.

include/lm/token_list.h
```diff
@@ -0,0 +1,116 @@
...
+
+    /**
+     * Remove the last token
+     */
+    void pop_back();
+
+    /**
+     * @return the underlying container of term_id tokens
+     */
+    const std::vector<term_id> tokens() const;
```
skystrife: `&`. `&` very important.

I missed this until I profiled.

smassung: Wow. Just fixed that.

PR #113, opened 2015-09-14
Filesystem improvements

skystrife:
## Motivation

Currently, we have calls directly to the system's `rm` executable in our unit tests. This is... obviously ugly and not portable. However, since `std::experimental::filesystem` isn't available yet on even our most up-to-date platforms (it's estimated that GCC 5.3.x and Clang 3.8.x will ship with it), we need to turn to something to act as a shim on our supported platforms.

* One option is to write that shim ourselves. That's probably not too hard on UNIX-like platforms, but I don't really want to figure out how to do it properly in Windows. I'm sure I could if sufficiently motivated, though.

* Another option is to use `boost::filesystem`, which is itself the basis for `std::experimental::filesystem`. The API would be familiar, and transitioning would be easy.

  ...but then we depend on Boost. It's one of the not-header-only libraries, and extracting it pulls in a lot of extra stuff as it is.

* The third option I've found is this: a library called [STLsoft](http://www.stlsoft.org/) (whose website does seem scarily out of date, but the software appears to be [actively maintained](https://github.com/meta-toolkit/meta-stlsoft/blob/master/CHANGES.txt#L5)) has a filesystem component, has a compatible license (BSD), and is header only! I tried messing with it and was actually fairly successful. It needed only a few very minor tweaks to silence compiler warnings and get things working on OS X. Those changes have been made in [a separate repository added as a submodule](https://github.com/meta-toolkit/meta-stlsoft) in this PR.

## Realization

This set of commits:

1. Moves `meta::filesystem` from the `meta-util` library to the `meta-io` library (as there's now an object file for it).

2. Adds two different implementations of the subset of file manipulation routines currently in use: one that uses `platformstl` from `stlsoft`, and one that uses `std::experimental::filesystem`. The `CMakeLists.txt` checks for the presence of `std::experimental::filesystem` and conditionally compiles one of these two versions.

  (The `std::experimental::filesystem` implementation is, of course, untested since I don't have a stdlib that actually implements it yet, but in principle it should just work once we do.)

3. Adds our fork of STLsoft as a submodule.

4. Updates all of the unit tests to use `meta::filesystem::remove_all()` instead of `system("rm -rf something")`

Curious what your thoughts were. There was a bit of work put into this, but I'm not going to be upset if this isn't merged because of the added dependency on some random library I found. Writing our own shim is certainly doable, especially since we really only need a small fraction of what the STLsoft library is providing. However, I think the reality might be that we never write it because, well, Windows, in which case this is a reasonable solution that can easily be removed once `std::experimental::filesystem` support is more widespread.


smassung:
I think this is a really minor dependency since
- it's header-only
- it's simply a submodule that is basically transparent to the user

Let's merge it. Once `std::experimental::filesystem` is out, we can just remove the submodule.

Issue #48, opened 2014-11-20
Add ability to load classifiers from model files


skystrife:
Currently some of the classifiers save model files (like `sgd`), but they don't have a way of initializing from a pre-trained model. For sgd specifically, the constructor calls `reset()` immediately.

We should add another constructor that allows for loading a pre-trained classifier from its model file. We'll need to change the way that the model file is stored most likely, but this should be a relatively easy change (for `sgd` anyway).

smassung:
Added in #114

PR #114, opened 2015-09-21
Refactor classify to use a more flexible dataset class

skystrife:
This is a major refactoring of the `meta::classify` library.

This introduces the `meta::learn` namespace, which will have more things in it as we move forward. The main motivation for this refactoring is to separate out the classifiers from the index objects to better support classification on new data. This also sets the groundwork for easily adding regression methods to the toolkit.

All classifiers now support saving/loading to/from disk, addressing issue #48. This is done via two factories (which users can be oblivious to) and two factory methods:
* `make_classifier()` for constructing **and** training a model in one go, and
* `load_classifier()` for loading a classifier from a `std::istream`. Since `istream` is pretty generic, this could be pretty much anything (and pushes the decision to gzip model files out to users instead of internally to the library).

This also adds quite a few more unit tests for the classifiers (to exercise the aforementioned save/load logic), and since the `dataset_view` class abstracts out data shuffling nicely, we achieve deterministic unit test behavior now (at least, the versions of clang and gcc that I've tried both agree on randomness now).

Essentially, there are now two important class hierarchies:

* `dataset` defines an in-memory collection of `feature_vectors` (pairs of `feature_id` and `double`).
* `dataset_view` is a lightweight wrapper over a `dataset`, supporting the commonly needed operations `shuffle()` and `rotate()`. It provides iterators and `begin()`/`end()`, which allow for iteration over the `instance` objects contained in a `dataset` in the order prescribed by the `dataset_view`. These objects are (somewhat) lightweight to create (though they do internally have a vector of integers internally, one for each referenced `instance`).

`datasets` actually store the data; `dataset_views` just view it.

There are some derived classes of both `dataset` and `dataset_view` to provide more functionality (e.g. for multiclass classification it is handy to know the number of classes in advance, and to have an easy mapping from `class_label` <--> `label_id`), but the same idea applies. Labels will be stored in derived classes of `dataset`. A simple helper `labeled_dataset<T>` is provided, which will likely be all we need for regression.

This is **such** a large refactoring that I'm **sure** I've missed something, somewhere. Looking for a review, both at a superficial (documentation) level as well as at the code level (are there better ways of doing things? is the design of this OK? etc.).


include/classify/binary_dataset_view.h
```diff
@@ -0,0 +1,89 @@
...
+
+namespace meta
+{
+namespace classify
+{
+
+using binary_dataset = learn::labeled_dataset<bool>;
+
+/**
+ * A non-owning view of a dataset with binary class albels. This is
```
smassung: labels

include/classify/classifier/winnow.h
```diff
@@ -52,24 +52,27 @@ class winnow : public classifier
      * Constructs a winnow classifier with the given multiplier,
      * error threshold, and maximum iterations.
      *
-     * @param idx The index to run the classifier on
+     * This constructor the winnow on the given training documents.
+     * Maintains a set of weight vectors \f$w_1,\ldots,w_K\f$ where \f$K\f$
```
smassung: constructor -> constructs?

include/classify/multiclass_dataset.h
```diff
@@ -0,0 +1,189 @@
...
+{
+  public:
+    using class_label_map = util::invertible_map<label_type, label_id>;
+    using class_label_iterator = class_label_map::iterator;
+
+    /**
+     * Creates an in-memory dataset from a forward_index. This loads the
+     * **entire index** into memory, so you should only use this on small
+     * datasets.
+     */
```
smassung: Can you give a pointer to which type to use if you don't think the dataset will fit in memory?

include/classify/multiclass_dataset_view.h
```diff
@@ -0,0 +1,116 @@
...
+    multiclass_dataset_view create_even_split() const
+    {
+        LOG(info) << "Creating an even split of class labels" << ENDLG;
+
+        using indices_type = std::vector<size_type>;
+
+        // partition by class label
+        std::unordered_map<class_label, indices_type> partitioned;
+        using value_type = decltype(partitioned)::value_type;
+        for (auto it = begin(), end = this->end(); it != end; ++it)
```
smassung: why `begin()` but `this->end()` ?

include/index/make_index.h
```diff
@@ -89,7 +89,7 @@ std::shared_ptr<Index> make_index(const cpptoml::table& config, Args&&... args)
         config, std::forward<Args>(args)...);
 
     // if index has already been made, load it
-    if (filesystem::make_directory(idx->index_name()) && idx->valid())
+    if (!filesystem::make_directory(idx->index_name()) && idx->valid())
```
smassung: wait what? was this a bug?

include/learn/dataset.h
```diff
@@ -0,0 +1,265 @@
...
+    /**
+     * Creates an in-memory dataset from a forward_index and a range of
+     * doc_ids, represented as iterators.
+     */
+    template <class ForwardIterator>
+    dataset(std::shared_ptr<index::forward_index> idx, ForwardIterator begin,
+            ForwardIterator end)
+        : total_features_{idx->unique_terms()}
+    {
+        uint64_t size = std::distance(begin, end);
```
smassung: should `uint64_t` technically be `size_type` here?

include/learn/dataset.h
```diff
@@ -0,0 +1,265 @@
...
+     * an inverted index isn't possible. This ctor is mainly for use with
+     * the knn classifier. The id field of the instance_types stored within
+     * the dataset is a document_id.
+     */
+    template <class ForwardIterator>
+    dataset(std::shared_ptr<index::inverted_index> idx, ForwardIterator begin,
+            ForwardIterator end)
+        : total_features_{idx->unique_terms()}
+    {
+        uint64_t size = std::distance(begin, end);
```
smassung: and possibly the same with these other `uint64_t`s

include/learn/dataset_view.h
```diff
@@ -0,0 +1,230 @@
...
+    // subset constructor
+    dataset_view(const dataset_view& dv, iterator first, iterator last)
+        : dset_{dv.dset_}, rng_{dv.rng_}
+    {
+        indices_.reserve(std::distance(first, last));
+        for (; first != last; ++first)
+            indices_.emplace_back(first.index());
+    }
+
+    void shuffle()
```
smassung: should the implementation of this function be part of the `random` namespace (e.g. taking two `random_access_iterator`s)? I can see myself wanting to use it to shuffle other things deterministically.

src/classify/classifier/nearest_centroid.cpp
```diff
@@ -75,22 +112,23 @@ class_label nearest_centroid::classify(doc_id d_id)
     return best_label;
 }
 
+template <class ForwardIterator>
 double nearest_centroid::cosine_sim(
-    const std::vector<std::pair<term_id, double>>& doc,
-    const std::unordered_map<term_id, double>& centroid)
+    ForwardIterator begin, ForwardIterator end,
+    const std::unordered_map<term_id, double>& centroid) const
```
smassung: maybe I should switch this to `sparse_vector` so we can get a faster `cosine_sim`...

src/classify/classifier/svm_wrapper.cpp
```diff
@@ -90,62 +171,30 @@ confusion_matrix svm_wrapper::test(const std::vector<doc_id>& docs)
...
-        std::getline(in, str_val);
-        uint32_t value = std::stoul(str_val);
-        class_label predicted = idx_->class_label_from_id(label_id{value});
-        class_label actual = idx_->label(d_id);
-        matrix.add(predicted, actual);
+        std::ifstream in("svm-predicted");
+        std::string str_val;
+        for (const auto& instance : docs)
+        {
+            // we can assume that the number of lines in the file is equal to
```
smassung: derp format

src/classify/classifier/svm_wrapper.cpp
```diff
@@ -90,62 +171,30 @@ confusion_matrix svm_wrapper::test(const std::vector<doc_id>& docs)
...
-    for (auto& d_id : docs)
     {
-        // we can assume that the number of lines in the file is equal to the
-        // number of testing documents
-        std::getline(in, str_val);
-        uint32_t value = std::stoul(str_val);
-        class_label predicted = idx_->class_label_from_id(label_id{value});
-        class_label actual = idx_->label(d_id);
-        matrix.add(predicted, actual);
+        std::ifstream in("svm-predicted");
```
smassung: braces?

src/classify/kernel/radial_basis.cpp
```diff
@@ -0,0 +1,85 @@
...
+            ++first_it;
+        }
+        else
+        {
+            dist += second_it->second * second_it->second;
+            ++second_it;
+        }
+    }
+
+    for (; first_it != first_end; ++first_it)
```
smassung: Can't you do these two for loops in the while loop if you change the termination condition? This is what I was thinking of doing for cosine sim

include/util/random.h
```diff
@@ -0,0 +1,105 @@
...
+ */
+
+#include <cstdint>
+#include <functional>
+#include <limits>
+#include <random>
+
+namespace meta
+{
+namespace random
```
smassung: namespace comment here or meta.h

smassung:
I think this refactoring is awesome!

General comments:
- should we specify version numbers of the files saved by the classifiers (and other things)? I remember we potentially had some issue with this with the CRF files
- the unit tests take about twice as long as they originally did, I'm assuming because of all the fileio, and that you have to test both creating and reading

include/learn/dataset.h
skystrife: Needs to be `uint64_t` for `printing::progress` ctor.

include/learn/dataset.h
skystrife: Same as above.

include/classify/binary_dataset_view.h
skystrife: Fixed in 27cd029.

include/classify/classifier/winnow.h
skystrife: Fixed in 27cd029.

include/classify/multiclass_dataset.h
skystrife: So you'd still use this *class*, but not that specific *constructor*.

Fixed in 7c81b64.

include/classify/multiclass_dataset_view.h
skystrife: Yep. I can change the name of the `end` variable, though, if you want.

include/index/make_index.h
skystrife: I think so. Probably introduced with the `filesystem` refactoring. `filesystem::make_directory` returns true if it actually *did* make the directory, and false otherwise (e.g. it already exists).

include/learn/dataset_view.h
skystrife: Agreed. Fixed in 0001be5.

src/classify/classifier/nearest_centroid.cpp
skystrife: The centroids? Maybe.

src/classify/classifier/svm_wrapper.cpp
skystrife: Fixed in 27cd029.

src/classify/classifier/svm_wrapper.cpp
skystrife: Fixed this (and several others of this; I think this code might be one of the older parts of MetA!) in a3a1f5c.

src/classify/kernel/radial_basis.cpp
skystrife: I don't think that will make it any more clear, since you'd need to tack on additional checks to the branches of the if/else chain to ensure that e.g. `first_it` or `second_it` is even valid to look at.

include/util/random.h
skystrife: Fixed in 0001be5.

skystrife:
- Re: model versioning---that's probably the *right* thing to do, but I don't feel super strongly about it for these, since we don't distribute any pre-trained classifier models with the releases. If you want me to add it, I can. I'll just have save write a 0 after the id and before the data and have the `istream&` ctor read and check that it sees a zero and throw otherwise.

- A quick sanity check shows that actually the largest contributor is the `one_vs_one` unit tests, which didn't exist before. If I comment those out on my desktop the running time goes down to approx. 6.79 sec from 12.68 sec.

include/classify/multiclass_dataset_view.h
smassung: No, not a big deal. Just made me pause for a second.

smassung:
Both responses are perfectly reasonable to me---it didn't occur to me that we don't distribute trained classifier models, and the unit test timing makes sense as well.

PR #116, opened 2015-10-21
Silence Clang conversion warnings

skystrife:
This also adds -Wconversion to the clang builds. GCC is still to noisy to bother with.

### DIFF
--- a/include/utf/utf.h
+++ b/include/utf/utf.h
@@ -26,15 +26,20 @@ namespace utf
  * @param dest The string to append the codepoint to
  * @param codepoint The UTF-32 codepoint to append
  */
-inline void utf8_append_codepoint(std::string& dest, uint32_t codepoint)
+inline void utf8_append_codepoint(std::string& dest, UChar32 codepoint)
 {
     std::array<uint8_t, U8_MAX_LENGTH> buf;
     int32_t len = 0;
     UBool err = FALSE;
+    // ICU has some conversions within this macro, which we can't control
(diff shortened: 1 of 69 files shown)


include/utf/utf.h
```diff
@@ -26,15 +26,20 @@ namespace utf
  * @param dest The string to append the codepoint to
  * @param codepoint The UTF-32 codepoint to append
  */
-inline void utf8_append_codepoint(std::string& dest, uint32_t codepoint)
+inline void utf8_append_codepoint(std::string& dest, UChar32 codepoint)
```
smassung: Would users ever have to call this function ? (Wondering about converting to `UChar32`)

src/parser/tools/parser_test.cpp
```diff
@@ -92,8 +92,8 @@ int main(int argc, char** argv)
         return 1;
     }
 
-    std::string path = *prefix + "/" + *treebank + "/treebank-3/parsed/mrg/"
-                       + *corpus;
+    std::string path
+        = *prefix + "/" + *treebank + "/treebank-3/parsed/mrg/" + *corpus;
```
smassung: Will building paths like this be compatible on Windows? (and all the other places that we do this)

Maybe we need a `filesystem::join_path`.

smassung: (After some Googling it seems that it's probably fine)

smassung:
:+1:

PR #120, opened 2015-11-24
Ensure ICU >= 56.1 via ExternalProject_Add.

skystrife:
The procedure for detecting ICU is now as follows:

1. Determine the version of the system ICU, if it exists
2. If the system version is new enough, use it
3. Otherwise, download and build a static version of ICU and link against it instead

This should eliminate the need for people to even have ICU installed at all when building, and ensures that the exact same version of the Unicode standard is used for e.g. icu_tokenizer's boundary detection.


skystrife:
OK, I think this is actually done and works now.

CMakeLists.txt
```diff
@@ -26,8 +26,90 @@ include_directories(include/)
 
 set(CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/deps/findicu)
 find_package(Threads REQUIRED)
-find_package(ICU COMPONENTS data i18n uc REQUIRED)
-include_directories(SYSTEM ${ICU_INCLUDE_DIRS})
+find_package(ICU 56.1 COMPONENTS data i18n uc)
+
+if (NOT ICU_FOUND OR ICU_VERSION VERSION_LESS 56.1)
```
smassung: should we store this 56.1 in a variable?

src/test/inverted_index_test.cpp
```diff
@@ -82,7 +82,11 @@ void check_ceeaus_expected(Index& idx)
     doc_id id{0};
     while (in >> size >> unique)
     {
+        std::cout << "size(" << id << "): " << idx.doc_size(id) << " vs "
+                  << size << std::endl;
```
smassung: keeping these print statements for output during failing unit tests?

CMakeLists.txt
skystrife: Yeah, let me do that. I should probably start encapsulating parts of the main `CMakeLists.txt` since it's getting rather long, too... so let me do that and then I'll ping you again for review.

src/test/inverted_index_test.cpp
skystrife: Yeah.

skystrife:
There. It's now a function that lives in a separate file, so now there's only one reference to the minimum required version. I figure it's not worth being a variable if it's only specified once, but I can always change that if you want.

smassung:
:+1:

PR #121, opened 2015-12-15
Add probing hash tables

skystrife:
This PR extends the hashing support in meta to provide more generic support for

  1.  probing-based hash sets, and
  2.  probing-based hash tables

with a variety of optimizations for specific key and value data types (for example: placing key/value data inside the table itself or externally, caching hash values, ...). While these tables currently only support insert and find operations (read: no deletions), they offer significant performance advantages over something like `std::unordered_X` in many scenarios.

This also comes with an additional feature in the form of support for aligned allocation. The `util::aligned_allocator<T, Size>` class template provides a STL-compatible allocator that will ensure that allocations are guaranteed to be aligned on at least `Size` byte boundaries, allowing for allocations in e.g. `std::vector<T>` to be over-aligned relative to the data type `T`. This is potentially a performance improvement that can allow SIMD optimizations without having to generate prelude and postlude loops to ensure alignment, etc. In this case, it allows the `probing::binary` strategy to be efficient by ensuring the hash table is properly aligned to a cache line (64 bytes on most processors today).

A template alias `util::aligned_vector<T>` has been added that represents a `std::vector<T>` that is guaranteed to be aligned on 64 byte boundaries. One can also use `util::aligned_vector<T, Size>` to align on different boundaries, if desired.


include/hashing/probing.h
```diff
@@ -0,0 +1,122 @@
...
+        : hash_{hash}, capacity_{capacity}, step_{0}
+    {
+        hash_ &= (capacity_ - 1);
+    }
+
+    /**
+     * @note This strategy only will work for power-of-2 capacities!
+     * @return the next index to probe in the table
+     */
+    uint64_t probe()
```
smassung: Do you think it's worth adding the discarding loop like in `binary_probing` so this probing strategy is at least able to run on non-power-of-2 sizes?

smassung:
Seems good to me, especially since we've been using it for a few weeks.

include/hashing/probing.h
```diff
@@ -0,0 +1,122 @@
...
+        hash_ %= capacity;
+    }
+
+    /**
+     * @return the next index to probe in the table
+     */
+    uint64_t probe()
+    {
+        // discard hashes that fall off of the table
+        for (; (hash_ ^ step_) >= capacity_; ++step_)
```
smassung: There might be some formula to avoid looping, as we've discussed.

smassung:
Merging; comments are minor and can be addressed later if necessary.

PR #122, opened 2015-12-27
bandit unit tests and refactoring

smassung:
The main contribution of this PR is to replace the existing unit test framework with [bandit unit tests](http://banditcpp.org/). Through this refactor, the existing unit tests are simplified (resulting in a good number of deletions) and more unit tests and test cases have been added.

While examining the entire code base again, some other maintenance changes were performed, such as
- removing unused includes
- making sure the correct constructor is called in `character_tokenize` (bugfix)
- removing outdated cmake variable `META_ICU_NO_TEMP_SUBSTRING`
- removing `binary_file_{reader,writer}`
- changing id of `normalizer` to `english-normalizer`
- removing `io::parser` and replacing with `std::fstream`
- adding various other error checking that was found useful when writing the test cases

There are also a few hashing-related modifications related to `murmur_hash` and `binary_probing` that were added into this branch.


skystrife:
:+1:

Issue #79, opened 2015-04-11
Move includes into a meta/ folder


skystrife:
If we think about eventually having meta in a package manager, it makes a lot of sense to isolate our headers into their own subdirectory, e.g. `/usr/include/meta/*`.

We'll need to change all of our internal includes to prefix `meta/`...

skystrife:
Fixed via b99d90a.

PR #123, opened 2015-12-28
#79 move includes into meta/ folder

smassung:
The main directory structure is now
- include/meta/
- src/
- tests/

All MeTA includes are now prefixed with `#include "meta/..."`.


skystrife:
I'm looking at the Appveyor failure now. Pretty sure it's unrelated.

PR #132, opened 2016-02-07
add zlib dependency to setup guide

smassung:
Assuming this seems fine, let's merge develop into master after this and before merging the other PR #131 . I'll make the same changes for meta-toolkit.org.


skystrife:
LGTM

PR #133, opened 2016-02-12
Add support for word embeddings

skystrife:
This PR adds the GloVe algorithm for training word embeddings and a library class `word_embeddings` for loading and querying trained embeddings.

To facilitate returning word embeddings, a simple `util::array_view` class was added. This class is a non-owning reference to an array (or part of one), enabling embedding vectors to be returned from the model without necessitating a heap allocation. Users can, of course, always copy the embedding into a `std::vector` via `std::vector<double> vec{av.begin(), av.end()};`.

To facilitate the demo application, and interaction with the embedding vectors in general, a simple vector math library has been added in `include/meta/math`. This adds overloads for `util::array_view` and `std::vector` for common math operations like +, -, /, and *. This file also contains definitions for computing the L1 and L2-norms. The overloads all live in the namespace `meta::math::operators`, which needs to be brought into scope before the overloads can be resolved via `using namespace meta::math::operators;`.

This also fixes a performance problem with `printing::progress` that was discovered in the process of building the embedding learning tools. The new version improved the performance of the first shuffle pass in the `glove` tool by 3x.


smassung:
Very cool.

The AppVeyor failure is fixed by some of my changes in another branch.

Future work
- write tutorial for word embeddings

Potential future work
- divide vector<int> by double
- replace vocab with array_view of hashed strings
- use stream from an analyzer as input to GloVe

PR #135, opened 2016-02-27
Add tokenize function to forward index

smassung:
PR for sanity check. Also breaks `learn::instance` into its own class.

### DIFF
--- a/tests/forward_index_test.cpp
+++ b/tests/forward_index_test.cpp
@@ -191,11 +191,18 @@ go_bandit([]() {
             corpus::document doc;
             doc.content(text);
             auto fvector = idx->tokenize(doc);
-            AssertThat(fvector.at(term_id{0}), Equals(1));
-            AssertThat(fvector.at(term_id{1}), Equals(1));
-            AssertThat(fvector.at(term_id{276}), Equals(1));
-            AssertThat(fvector.at(term_id{3343}), Equals(2));
-            AssertThat(fvector.at(term_id{3731}), Equals(1));
+
+            auto begin_sent = idx->get_term_id("<s>");
+            auto end_sent = idx->get_term_id("</s>");
+            auto bad = idx->get_term_id("bad");
+            auto smoke = idx->get_term_id("smoke");
+            auto think = idx->get_term_id("think");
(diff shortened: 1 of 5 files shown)


tests/forward_index_test.cpp
```diff
@@ -183,6 +183,21 @@ go_bandit([]() {
...
+            std::string text{"I think smoking smoking bad."};
+            corpus::document doc;
+            doc.content(text);
+            auto fvector = idx->tokenize(doc);
+            AssertThat(fvector.at(term_id{0}), Equals(1));
+            AssertThat(fvector.at(term_id{1}), Equals(1));
+            AssertThat(fvector.at(term_id{276}), Equals(1));
+            AssertThat(fvector.at(term_id{3343}), Equals(2));
+            AssertThat(fvector.at(term_id{3731}), Equals(1));
+            AssertThat(fvector.at(term_id{2}), Equals(0));
```
skystrife: This test seems very fragile. Could we look up the term id for the specific terms via `disk_index::get_term_id()` instead of using hard-coded term ids?

smassung: Yes, definitely.

tests/forward_index_test.cpp
```diff
@@ -183,6 +183,28 @@ go_bandit([]() {
...
+            auto bad = idx->get_term_id("bad");
+            auto smoke = idx->get_term_id("smoke");
+            auto think = idx->get_term_id("think");
+
+            AssertThat(fvector.at(begin_sent), Equals(1));
+            AssertThat(fvector.at(end_sent), Equals(1));
+            AssertThat(fvector.at(bad), Equals(1));
+            AssertThat(fvector.at(smoke), Equals(2));
+            AssertThat(fvector.at(think), Equals(1));
+            AssertThat(fvector.at(term_id{2}), Equals(0));
```
skystrife: This one, too. 2 could end up being one of the term ids actually in the sentence, especially if (when?) we switch the vocabulary map implementation.

smassung: Fair enough, also changed.

PR #137, opened 2016-03-04
Add minimal perfect hashing

skystrife:
This PR contains a number of things, but they are all in support of adding minimal perfect hash function construction to the toolkit.

- `util::multiway_merge` has been made more generic so that users can specify their own sorting and merging criteria.
- `meta::succinct` contains a library of a few simple (quasi-)succinct rank/select data structures. These answer questions on sparse or dense bit vectors of the form "where is the next 0/1" (select) and "how many 0/1s occurred before this bit position" (rank). These data structures and their builders are all decently unit tested.
- `io::mifstream` and `io::mofstream` have been added as workarounds for the lack of `std::fstream` move constructors on older compilers (GCC < 5).
- `hashing::perfect_hash_builder<K>` is a builder to construct minimal perfect hash functions from a set of (deduplicated) keys. `hashing::perfect_hash<K>` is a query class for the hash functions built. The code here uses a modified version of the CHD (hash, displace, and compress) algorithm for generating minimal perfect hash functions using a disk-streaming algorithm. Empirically, our minimal hash functions come out to about ~2.7 bits/key of overhead with the current implementation.

At this point, only the hash *function* generation exists (so there are no maps/sets), and it's expected that this should support building a few other things (like a better `vocabulary_map` and language model implementation).


smassung:
LGTM. Has good comments, code structure, unit tests.

PR #148, opened 2016-07-12
fix a reference error.

yxkemiya:
`http://` prefix missed in the reference url for `deprecated in C++14`, so the redirection is wrong and it will lead to 404.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -102,8 +102,6 @@ CompilerKludges()
 
 if(NOT META_HAS_EXPERIMENTAL_FILESYSTEM)
   target_include_directories(meta-definitions SYSTEM INTERFACE ${PROJECT_SOURCE_DIR}/deps/meta-stlsoft/include)
-else()
-  target_compile_options(meta-definitions INTERFACE -lc++experimental)
 endif()
 
 target_link_libraries(meta-definitions INTERFACE compiler-kludges)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -102,6 +102,8 @@ CompilerKludges()
 
 if(NOT META_HAS_EXPERIMENTAL_FILESYSTEM)
   target_include_directories(meta-definitions SYSTEM INTERFACE ${PROJECT_SOURCE_DIR}/deps/meta-stlsoft/include)
+else()
+  target_compile_options(meta-definitions INTERFACE -lc++experimental)
 endif()
 
 target_link_libraries(meta-definitions INTERFACE compiler-kludges)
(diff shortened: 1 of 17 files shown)


skystrife:
Good stuff in here, but:

1. There's too much in this one PR. Each separate issue should be a separate PR.
2. I appreciate the effort in fixing the build, but I think a better solution is for us to not build against LLVM's trunk, but instead use a specific tagged version. This should make CI more stable.

  As far as supporting libc++'s `experimental::filesystem` implementation, I'll wait until the release of LLVM 3.9.0 before diving too deep into that.
3. I'm not sure I fully understand the bugfix in the last commit. Can you make a separate PR that describes the bug (with reproduction instructions) along with your fix? (Make sure you follow the style guide, too: I'm seeing tabs.)

The fix for `STYLE.md` is fine, though, so I've cherry-picked that into the `develop` branch as a58978e4ae7eefe8d21c6c0fd76804dc4dd2ca11.

yxkemiya:
@skystrife  I'm so sorry that I have do so many commit in one PR, but all commit except the first one is try to fix the build for supporting libc++'s `experimental::filesystem`. 

The last commit is also a bug fix for supporting libc++'s `experimental::filesystem`. I test the `experimental::filesystem` feature on my Ubuntu by using the following code which is modified from [here](http://en.cppreference.com/w/cpp/experimental/fs/create_directory):
```
#include <iostream>
#include <fstream>
#include <cstdlib>
#include <experimental/filesystem>
namespace fs = std::experimental::filesystem;
 
int main()
{
    //fs::create_directory("sandbox/1/2/a"); //This will throw std::experimental::filesystem::v1::filesystem_error if the parent directory `sandbox/1/2` is not exists.
    fs::create_directories("sandbox/1/2/a");
    //std::cout << std::boolalpha << fs::create_directories("sandbox/1/2/a") << std::endl; // It will return false if the directory `sandbox/1/2/a` exists already.
    fs::remove_all("sandbox");
}
```
So
1.  The modification in `src/io/filesystem.cpp` is trying to use `make_directories` instead of `make_directory` to avoid the `std::experimental::filesystem::v1::filesystem_error` if the input directory is a subdir of a nonexistent directory.
2.  The modification in `src/index/forward_index.cpp` and `src/index/inverted_index.cpp` means that, if the index directory is already exists, the exception `throw exception{"Unable to create index directory: " + index_name()};` is misleading, and the function is to create index, I think the index directory should not exists before trying to create directory, so I try to delete the index directory if it's already exists.

If you want to support libc++ `experimental::filesystem` implementation until the release of LLVM 3.9.0, the other commits will be useless, and hope it will give you some help when you want to support libc++ `experimental::filesystem` implementation, but if you would like to accept what I have done on `experimental::filesystem`, please let me know and I will try to do another PR.

Thanks a lot for accepting my PR and hope my poor english will not bother you so much.

Issue #149, opened 2016-08-01
Feature selection on dataset_view instead of forward_index


siddshuk:
feature_selector requires a forward_index to be passed in to the constructor. But the forward_index is built on the entire dataset, which includes training and test data.

Feature selection should be done on training data separately without touching the test data to avoid optimistically biased performance estimates during cross-validation.

Is it possible to allow feature selection to be done on a dataset_view (view for the training portion of the dataset in this case)?

Thanks.

Best Regards,
Sid

smassung:
Great question. Unfortunately, getting feature selection integrated completely with indexes/datasets is still on the to do list as issue #111 

Right now the feature selection can really only show stats for a particular corpus, e.g. by using`./feature-summary` (https://github.com/meta-toolkit/meta/blob/master/src/features/tools/feature_summary.cpp)

As always though, if anyone is interested in implementing this, we would gladly accept a pull request :smiley:

smassung:
If you need this functionality immediately (and are just using unigram words), a super hacky way to accomplish this would be to:
- separate out your training data and test data as different corpora
- run feature selection on the training data to get a list of the top k words
- learn a classifier on the training data with a list filter on the k words using method=accept (only analyze words from that list of top k)
- save your classifier model that was built on the training corpus
- load your classifier model and run it on the test corpus

I think that should work...

siddshuk:
Thanks for the quick response. I'll give it a try.

skystrife:
Fixed by PR #150.

Issue #111, opened 2015-09-12
create selected_index using feature selection


smassung:
Create an index wrapper that uses the feature selection algorithms to determine whether or not to return certain terms from the index back to the user/application.

skystrife:
Based on the way `meta-classify` now works, I think this is a ctor for `learn::dataset` that takes an index and a feature selector and selectively loads in only the active features into memory.

skystrife:
So I set about starting to do this and realized that there's sort of a big choice to make.

First, the easy way to do this: just take an existing `labeled_dataset` and `feature_selector`. Iterate through the `dataset` and filter each `feature_vector` to only contain features that are marked selected via `selector.selected(feat_id)`. This will *work*, but will result in classification models that can end up being way larger than they need to be due to containing weights for lots of now unused feature ids.

An alternative is to somehow "condense" the feature id space by performing a relabeling by pretending all non-selected `feature_id`s don't exist. As a simple example, if we had feature ids [0, 1, 2, 3, 4, 5] but only selected [2, 4, 5], we would relabel as 2 -> 0, 4 -> 1, and 5 -> 2.

The problem now is that the `feature_ids` no longer correspond to `term_ids` from the original `forward_index`, so you can't go look up the meaning of weights as easily (or at all) anymore. So we'd likely need to *also* produce a backward mapping from new id -> old id.

Fortunately, I think we can do these efficiently using stuff from the `meta-succinct` library (in particular, the `sarray`). Using `sarray_select`, we can answer questions like "give me the index of the i-th one in a bit vector". In our case, the "bit vector" is the selected features, where 0 indicates unselected and 1 indicates selected, and the query `select(new_id)` would return `old_id`. Using `sarray_rank`, we can answer questions like "how many ones are before the i-th position in a bit vector". In our case, `rank(old_id)` (where `old_id` is indeed a selected feature) would return `new_id`.

Perhaps unfortunately, all of the stuff in `succinct` is currently designed around keeping storage in memory-mapped files (so they can be easily persisted to disk). That would be totally unnecessary here, and it'd be confusing to have to give a folder prefix to create a new feature-selected dataset. So in order to actually do this, I'd need to do a (probably beneficial anyway) refactoring of `meta-succinct` to support non-persisted storage.

(...sorry, this is turning into a stream-of-consciousness note.)

[...]

PR #150, opened 2016-08-17
Feature selection support for binary/multiclass/regression datasets and dataset views

siddshuk:
Updated feature selection so as it can be performed on labeled (i.e, binary or multiclass or regression) datasets and dataset views. Also, added a constructor in multiclass_dataset which takes in a featurizer function as a parameter.

As shown below, these changes allow us to run feature selection on a portion of the data (such as the training data) and then create a new dataset consisting of only the selected features.

```
auto selector = features::make_selector(*config, training_vw);

uint64_t total_features_selected = 20;

classify::multiclass_dataset feature_selected_dset
(
   training_vw.begin(), training_vw.end(), total_features_selected, 
   [&](const learn::instance & inst)
   {
        learn::feature_vector inst_weights;
        for(const auto& w : inst.weights)
        {
            if(selector->selected(w.first))
               inst_weights.emplace_back(w.first, w.second);
        }

        return inst_weights;
   },
   [&](const learn::instance inst)
   {
       return training_vw.label(inst);
   }
);
```

This is in reference to issues #149 and #111.


smassung:
Thanks for the PR! Just wanted to let you know that we will take a look at this over the next few days.

siddshuk:
Great, thanks! I look forward to hearing your feedback.

skystrife:
First of all, I'm very impressed with the code quality here. Very nice work. With some minor changes, I think this is very mergeable. :+1: 

A few quick things from my perspective:

1. It looks like there are still some tabs in the source files, and a few minor stylistic differences from the style guide. If possible, can you run a quick `clang-format` on the files you've touched? This should correct that and any other obvious style issues.

  (If you don't have access to `clang-format`, we can always do that for you before merging, so it's not a huge deal. But it's nice to just "press button receive merge". :wink:)
2. I *think* (@smassung, correct me if I'm wrong here) that the current feature selection code is likely only to work for multiclass (and maybe binary) datasets. The methods implemented are looking at statistics like `Pr(feature, class)` which really breaks down with regression datasets where the labels are going to be real numbers (and where you could in theory have as many "labels" as you do data points).

  (This is actually a good point: @smassung, there are feature selection methods that aren't covered under the current paradigm, like "give me only features with non-negligable variance". What's your position on supporting things like that?)

  Things will likely need tweaked more still to work for `binary_dataset_view`s. For example, `bdv.label(instance)` will return `bool`, not `class_label` like you're expecting here.

  For now, I'd recommend taking the path of least resistance and just focus on making things work on `multiclass_dataset_view`. We can mess around with it after that to ensure it works for `binary_dataset_view` once things are merged.
3. I'm noticing changes from using `label_id` to `class_label`. I think I understand the reasoning behind that (you need stable label identifiers?), but if you're going to be passing around `class_label` a lot I recommend passing via `const&` to avoid copies. `class_label` is just a `std::string`, after all, unlike the other `_id`s.

@smassung, can you look more closely at the actual changes made to the `feature_selector` for sanity checking, since you're more familiar with that part of the code?

siddshuk:
@skystrife Thanks for the feedback! I'm really glad to hear that you like the code quality.

1. As suggested, I ran clang-format on all touched files. Hopefully, that should fix all the formatting issues and make it easier for you to merge.

2. I updated the feature selection code to convert all label types to a class_label (using `std::stringstream`) when calculating the probabilities. I think that should fix the issues you mentioned with the binary and regression datasets.

3. Good catch! I updated the class_label parameters to be passed via `const&`.

skystrife:
1. Thanks!
2. This fix should work for `binary_dataset_view`, but not for `regression_dataset_view`. I just flat out don't think the existing feature selection framework we have will work for a regression dataset.

  Here's just a quick example: you have a dataset of people and you want to predict their weights, in pounds, given other features about the individual. There's no guarantee that a large number of people will share the *exact* same weight, so you'll get a huge number of classes (likely close to the number of examples) and the selection algorithm won't do what it's supposed to (since there is only potentially one or two examples for every "class" here).

  This isn't an issue caused by your proposed change, but rather caused by the feature selection setup in general. I'd suggest just removing support for `regression_dataset_view` since it just doesn't quite make sense with the existing framework. Figuring out how to get feature selection working for regression is a separate issue, I think, that requires a different approach than the classification-centric design that is implemented already.
3. Great, thanks!

smassung:
@siddshuk , @skystrife , thanks for the great discussion and sorry for the delay -- I was traveling but am back now.

@skystrife Re: 2 from both of your comments -- yes, the feature selection code was only designed to be used for multiclass classification. It was essentially just implementing some formulas described in section 2 [this paper](http://sci2s.ugr.es/keel/pdf/specific/articulo/zheng.pdf). I agree that we should worry about regression feature selection as a separate issue.

The PR looks excellent after the latest commits. I especially like the use of `meta::stats` to replace the manual probability calculations. The unit tests are also up to date. They are very simplistic, but they were like that originally.

Let's merge this. I will create a separate issue to remind us to get things working generally, specifically with regression.

Again, thanks to @siddshuk ! :+1:

siddshuk:
@skystrife Ah, I see your point regarding 2 now. Thanks for the helpful example. 👍 

@smassung I'm glad you like the changes as well.

@skystrife, @smassung, thanks again for the helpful feedback and for merging this PR!

PR #154, opened 2016-08-28
Add a minimal perfect hashing language model implementation

skystrife:
This PR contains code that coexists with the existing language model that implements a minimal perfect hashing-based language model using the perfect hashing framework added in a previous merge.

The LM's query interface is different from the existing language model and more closely models the interface provided by KenLM. It produces (on my limited test data) identical perplexity results, but is not particularly fast.

A few additional minor things are included:

* `std::vector<T>` can now be hashed using the `hash_append` framework.

* `std::vector<T>` can be read/written using `io::packed`.

* `io::packed` has been refactored to make adding packed reading/writing routines for user-defined types easier, and is similar in spirit to `hash_append`.

  Users can provide `packed_read` and `packed_write` functions in the same namespace scope as their type, and `io::packed::read` and `io::packed::write` will find them via ADL.

* All `chunk_iterators` for `util::multiway_merge` have been consolidated into a single template class. This reduces quite a bit of code duplication across all of the disk-based merge algorithms we have.

### DIFF
--- a/include/meta/succinct/darray.h
+++ b/include/meta/succinct/darray.h
@@ -304,8 +304,10 @@ class darray
         {
             using namespace darray_detail;
 
+#if DEBUG
             if (META_UNLIKELY(i >= num_ones))
                 throw std::out_of_range{"index out of range in select query"};
+#endif
 
             auto block_idx = i / ones_per_block;
             if (blocks[block_idx] < 0)
--- a/include/meta/succinct/darray.h
+++ b/include/meta/succinct/darray.h
@@ -10,7 +10,6 @@
 #ifndef META_SUCCINCT_DARRAY_H_
 #define META_SUCCINCT_DARRAY_H_
 
-#include <iostream>
 #include <fstream>
 #include "meta/io/binary.h"
 #include "meta/io/filesystem.h"
@@ -213,6 +212,7 @@ class darray_builder
             {
                 auto offset = static_cast<uint16_t>(current_block[i]
                                                     - current_block[0]);
+                assert(i == 0 || offset > 0);
                 io::write_binary(sub_blocks, offset);
             }
         }
@@ -291,7 +291,7 @@ class darray
         {
             using namespace darray_detail;
 
-            if (META_UNLIKELY(i > num_ones))
+            if (META_UNLIKELY(i >= num_ones))
                 throw std::out_of_range{"index out of range in select query"};
 
             auto block_idx = i / ones_per_block;
@@ -323,7 +323,7 @@ class darray
             {
                 auto popcount = broadword::popcount(word);
                 if (one_count < popcount)
-                  break;
+                    break;
                 one_count -= popcount;
                 word = WordReader{}(words[++word_idx]);
             }
--- a/src/lm/mph_language_model.cpp
+++ b/src/lm/mph_language_model.cpp
@@ -251,21 +251,134 @@ uint64_t build_from_arpa(const std::string& arpa_file,
 }
 }
 
+struct mph_language_model::impl
+{
+    using unigram_map_type = ngram_map<std::string>;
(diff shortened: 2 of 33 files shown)


include/meta/succinct/darray.h
```diff
@@ -303,8 +304,10 @@ class darray
         {
             using namespace darray_detail;
 
-            if (META_UNLIKELY(i > num_ones))
+#if DEBUG
+            if (META_UNLIKELY(i >= num_ones))
```
smassung: Did this fix an off-by-one error?

src/lm/mph_language_model.cpp
```diff
@@ -0,0 +1,411 @@
...
+
+namespace meta
+{
+namespace lm
+{
+
+namespace
+{
+template <class NGramHandler>
+void read_from_arpa(const std::string& filename, NGramHandler&& ngram_handler)
```
smassung: Is it worth trying to share this parsing code with the other LM?

src/lm/mph_language_model.cpp
```diff
@@ -0,0 +1,411 @@
...
+        auto time = common::time(
+            [&]() { order = build_from_arpa(*arpa_file, *prefix); });
+        LOG(info) << "Done. (" << time.count() << "ms)" << ENDLG;
+    }
+    LOG(info) << "Loading language model from binary files in: " << *prefix
+              << ENDLG;
+
+    if (!order)
+    {
+        for (uint64_t o = 0; filesystem::file_exists(
```
smassung: Minor: maybe use different variable name than `o`? Appears in impl constructor as well.

smassung:
Are you able to compare MPH speed vs the current LM speed?

skystrife:
Probably, but not without a bit of work. Their API for querying isn't the same, so I would need to modify one or the other to offer an equivalent query interface to the other. I think the overall query style of the mph-lm is better since it can reduce the number of redundant failing lookups done since the `lm_state` remembers how far the model backed off on the previous query. The existing lm doesn't retain this info and always starts looking up the full ngram each time.

smassung:
Cool. If there is a top-*k* function in MPH, then we should be able to replace the old LM. What do you think? The one in the current LM is very inefficient (scan over all words in vocab), so the MPH one wouldn't need to be performant...

smassung:
Or "replace old LM with MPH LM" could be a separate issue (which I could look at). I'd like to see `diff_analyzer`'s performance increase.

skystrife:
There isn't an efficient top-*k* implementation here, since in principle this is the same format as the original lm, just with a perfect hash function instead of a regular general-purpose one. Now, you could *make* an efficient top-*k* on top of the code here, but you would need additional processing and a separate lookup structure (like the postings list style thing we talked about).

skystrife:
OK, I've done some speed testing experiments. We should definitely keep the existing LM around based on these results.

| Model | Query speed | Model size |
|---|---|---|
| **mph** | 3.98 MiB/s | 1.6G |
| **lm** | 8.83 MiB/s | 3.4G |
| **kenlm trie** | 9.98 MiB/s | 1.6G |

So basically, the mph language model is a little more than 2.2x slower than the probing one, and is about 2.5x slower than KenLM's trie implementation.

The query code I used between the regular lm and the mph-lm was basically identical, so the huge performance hit comes from evaluating the perfect hash function. I dug deeper into that, and we lose most of the performance on the call to an `sarray_rank.rank(k)`. It seems as though the couple general-purpose hash functions invoked aren't slowing things down a lot. So if we can come up with a better performing sparse succinct rank structure, it should plug into the mph model very easily and result in a large speedup.

smassung:
I think this should be good to go now, and we just keep both implementations with the unified query code?

PR #155, opened 2016-09-17
Embedding analyzer

smassung:
This adds an `embedding_analyzer` class that simply represents a document as its averaged word embeddings.

Using the IMDB dataset and 2-fold CV, unigram words get ~88% accuracy and averaged gigaword 300d embeddings get ~80% accuracy (on balanced classes). The `embedding_analyzer` class can be considered a baseline document representation method and can be expanded/improved for future work.

The weirdness with this analyzer is how it converts the averaged vectors to features using the featurizer. If used with an inverted index, the features need to be `uint64_t`s. How the analyzer is written now can work with `inverted_index` by converting the small `double`s to positive 16 digit integers with `abs()`. However, we lose some information. If we keep them as signed doubles, we can actually increase to classification accuracy on IMDB (and presumably other datasets) to ~82%, but the data can't be stored in an inverted index.

What's your opinion on this? Is there some cute way to convert or store these signed doubles as positive integers without using `abs()`? I tried normalization like so:

```cpp
    auto max_elem = *std::max_element(features.begin(), features.end());
    auto min_elem = *std::min_element(features.begin(), features.end());
    uint64_t cur_dim = 0;
    for (const auto& f : features)
    {
        auto val = (f - min_elem) / (max_elem - min_elem) * 1e16;
        counts(std::to_string(cur_dim++), static_cast<uint64_t>(val));
    }
```

But that somehow loses important information and causes the classification accuracy to drop to ~61%. I can't use something like `exp` since that destroys the relative distances between the feature values.

### DIFF
--- a/include/meta/embeddings/analyzers/embedding_analyzer.h
+++ b/include/meta/embeddings/analyzers/embedding_analyzer.h
@@ -29,12 +29,8 @@ namespace analyzers
  * [[analyzers]]
  * method = "embedding" # this analyzer
  * filter = # use same filter type that embeddings were learned with
- *
- * [embeddings]
- * prefix = "path/to/learned/embeddings"
+ * prefix = "path/to/embedding/model/"
  * ~~~
- *
- * Optional config parameters: none.
  */
 class embedding_analyzer : public util::clonable<analyzer, embedding_analyzer>
 {
@@ -64,6 +60,12 @@ class embedding_analyzer : public util::clonable<analyzer, embedding_analyzer>
 
     /// Learned word embeddings
     std::shared_ptr<embeddings::word_embeddings> embeddings_;
+
+    /// Path to the embedding model files
+    std::string prefix_;
+
+    /// Storage for the aggregated word embeddings per document
+    std::vector<double> features_;
 };
 
 /**
--- a/include/meta/embeddings/analyzers/embedding_analyzer.h
+++ b/include/meta/embeddings/analyzers/embedding_analyzer.h
@@ -21,7 +21,9 @@ namespace analyzers
 {
 
 /**
- * Analyzes documents by averaging word embeddings for each token.
+ * Analyzes documents by averaging word embeddings for each token. This analyzer
+ * should only be used with forward_index since it stores double features
+ * values.
  *
  * Required config parameters:
  * ~~~toml
--- a/include/meta/embeddings/analyzers/embedding_analyzer.h
+++ b/include/meta/embeddings/analyzers/embedding_analyzer.h
@@ -22,8 +22,7 @@ namespace analyzers
 
 /**
  * Analyzes documents by averaging word embeddings for each token. This analyzer
- * should only be used with forward_index since it stores double features
- * values.
+ * should only be used with forward_index since it stores double feature values.
  *
  * Required config parameters:
  * ~~~toml
--- a/include/meta/embeddings/analyzers/embedding_analyzer.h
+++ b/include/meta/embeddings/analyzers/embedding_analyzer.h
@@ -1,0 +1,84 @@
+/**
+ * @file embedding_analyzer.h
+ * @author Sean Massung
+ *
+ * All files in META are released under the MIT license. For more details,
+ * consult the file LICENSE in the root of the project.
+ */
+
+#ifndef META_EMBEDDINGS_EMBEDDING_ANALYZER_H_
+#define META_EMBEDDINGS_EMBEDDING_ANALYZER_H_
+
+#include "meta/analyzers/analyzer.h"
+#include "meta/analyzers/analyzer_factory.h"
+#include "meta/embeddings/word_embeddings.h"
+#include "meta/util/clonable.h"
+#include <string>
+
+namespace meta
+{
+namespace analyzers
+{
+
+/**
+ * Analyzes documents by averaging word embeddings for each token.
+ *
+ * Required config parameters:
+ * ~~~toml
+ * [[analyzers]]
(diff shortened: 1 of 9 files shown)


skystrife:
I say just disallow use of this analyzer for `inverted_index`, as it's not clear why you would want to do so anyway (what would the scoring functions for retrieval even mean?).

There should be an exception thrown at runtime if someone tries to do this, so they should be protected from that potential foot-gun. Just leave a message in the class comment informing the user that it can only be used for `forward_index`.

include/meta/embeddings/analyzers/embedding_analyzer.h
```diff
@@ -0,0 +1,86 @@
...
+#include <string>
+
+namespace meta
+{
+namespace analyzers
+{
+
+/**
+ * Analyzes documents by averaging word embeddings for each token. This analyzer
+ * should only be used with forward_index since it stores double features
```
skystrife: s/features/feature

smassung:
Sounds good to me!

include/meta/embeddings/analyzers/embedding_analyzer.h
smassung: :+1:

src/embeddings/analyzers/embedding_analyzer.cpp
```diff
@@ -0,0 +1,75 @@
...
+    {
+        auto token = stream_->next();
+        features = features + embeddings_->at(token).v;
+        ++num_seen;
+    }
+
+    // average each feature and record it
+    uint64_t cur_dim = 0;
+    for (const auto& val : features)
+        counts(std::to_string(cur_dim++), val / num_seen);
```
skystrife: I think we might want to have a prefix here to avoid conflating unigram number tokens from a `ngram_word_analyzer` and the embedding dimension feature values if they are used together.

src/embeddings/analyzers/embedding_analyzer.cpp
```diff
@@ -0,0 +1,75 @@
...
+{
+    // nothing
+}
+
+void embedding_analyzer::tokenize(const corpus::document& doc,
+                                  featurizer& counts)
+{
+    using namespace math::operators;
+    stream_->set_content(get_content(doc));
+    std::vector<double> features(embeddings_->vector_size(), 0.0);
```
skystrife: I doubt this is worth it, but we could potentially avoid allocating space for this `vector` each time by storing it as a member variable, since `embeddings_->vector_size()` won't change for the lifetime of the analyzer.

src/embeddings/analyzers/embedding_analyzer.cpp
```diff
@@ -0,0 +1,75 @@
...
+        counts(std::to_string(cur_dim++), val / num_seen);
+}
+
+template <>
+std::unique_ptr<analyzer>
+make_analyzer<embedding_analyzer>(const cpptoml::table& global,
+                                  const cpptoml::table& config)
+{
+    auto filts = load_filters(global, config);
+    return make_unique<embedding_analyzer>(global, std::move(filts));
```
skystrife: For consistency, I think it would be better if the embeddings prefix was stored inside the `[[analyzer]]` block rather than pulled from the global configuration. The `ngram_pos_analyzer` and `tree_analyzer` both do this.

src/embeddings/analyzers/embedding_analyzer.cpp
smassung: good catch

src/embeddings/analyzers/embedding_analyzer.cpp
smassung: sounds good

src/embeddings/analyzers/embedding_analyzer.cpp
smassung: sure, might as well

src/embeddings/analyzers/embedding_analyzer.cpp
smassung: ooh, then you can include multiple embedding sources...

src/embeddings/analyzers/embedding_analyzer.cpp
```diff
@@ -0,0 +1,75 @@
...
+                                  featurizer& counts)
+{
+    using namespace math::operators;
+    stream_->set_content(get_content(doc));
+    std::vector<double> features(embeddings_->vector_size(), 0.0);
+    uint64_t num_seen = 0;
+    while (*stream_)
+    {
+        auto token = stream_->next();
+        features = features + embeddings_->at(token).v;
```
skystrife: And now that I think about it, I wonder if this would be faster:

```cpp
features = std::move(features) + embeddings_->at(token).v;
```

(since what we really want to say is `+=`, but we can't add that member function to `std::vector`.)

src/embeddings/analyzers/embedding_analyzer.cpp
skystrife: I didn't even consider that. Might need to be extra careful with the feature prefix, in that case. That's probably something problematic in the POS and tree analyzers, too.

smassung: exactly

src/embeddings/analyzers/embedding_analyzer.cpp
smassung: will add it, but it's already pretty fast!

skystrife:
I'm going to merge this. Just so we don't hate ourselves later, can you update the `CHANGELOG`?

PR #173, opened 2016-11-01
Add an unsupervised HMM implementation

skystrife:
This PR adds an unsupervised HMM implementation. This implementation supports HMMs with discrete observations (what is used most often) and sequence observations (useful for log mining applications). The forward-backward algorithm is implemented using both the scaling method and the log-space method. The scaling method is used by default, but the log-space method is useful for HMMs with sequence observations to avoid underflow issues when the output probabilities themselves are very small.

The EM algorithm to fit the HMM is parallelized.

There are a few minor enhancements that were added in this branch to support the research work that motivated this implementation; these are mentioned in the `CHANGELOG.md` modification.


include/meta/sequence/hmm/discrete_observations.h
```diff
@@ -0,0 +1,151 @@
...
+     */
+    template <class Generator>
+    discrete_observations(uint64_t num_states, uint64_t num_observations,
+                          Generator&& rng,
+                          stats::dirichlet<observation_type>&& prior)
+        : obs_dist_(num_states, prior)
+    {
+        for (auto& dist : obs_dist_)
+        {
+            for (observation_type o{0}; o < num_observations; ++o)
```
smassung: Can you replace `o` with `obs` or something?

include/meta/sequence/hmm/forward_backward.h
```diff
@@ -0,0 +1,385 @@
...
+namespace meta
+{
+namespace sequence
+{
+namespace hmm
+{
+
+/**
+ * Encapsulates the forward-backward algorithm using the scaling method
+ * from the original Rabiner paper.
```
smassung: Citation link would be helpful here (unless it's already somewhere else)

include/meta/sequence/hmm/forward_backward.h
```diff
@@ -0,0 +1,385 @@
...
+        {
+            scratch[i] = fwd.probability(seq.size() - 1, i);
+        }
+        counts.log_likelihood += log_sum_exp(scratch.begin(), scratch.end());
+    }
+};
+}
+}
+}
+#endif
```
smassung: I don't see an obvious way to consolidate the `logarithm` and `scaling` variants into the same code without introducing more complexity, so I guess we can just leave it as is.

include/meta/sequence/hmm/hmm.h
```diff
@@ -0,0 +1,303 @@
...
+            }
+            else
+            {
+                LOG(info) << "Log log_likelihood: " << log_likelihood << ENDLG;
+            }
+
+            if (old_ll > log_likelihood)
+            {
+                LOG(fatal) << "Log likelihood did not improve!" << ENDLG;
+                throw std::runtime_error{"Log likelihood did not improve"};
```
smassung: `hmm_exception`?

PR #174, opened 2016-11-11
Add KL-divergence based pseudo-relevance feedback ranker

skystrife:
### Main Changes
Add the KL-divergence retrieval function using pseudo-relevance feedback with the two-component mixture-model approach of Zhai and Lafferty, called `kl_divergence_prf`. This ranker internally can use any `language_model_ranker` subclass like `dirichlet_prior` or `jelinek_mercer` to perform the ranking of the feedback set and the result documents with respect to the modified query.

The EM algorithm used for the two-component mixture model is provided as the `index::feedback::unigram_mixture` free function and returns the feedback model.

This PR also breaks the ranker hierarchy into one more level. At the top we have `ranker`, which has a pure virtual function `rank()` that can be overridden to provide entirely custom ranking behavior. This is the class the KL-divergence method derives from, as we need to re-define what it means to rank documents (first retrieving a feedback set, then ranking documents with respect to an updated query).

Most of the time, however, you will want to derive from the second level `ranking_function`, which is what was called `ranker` before. This class provides a definition of `rank()` to perform document-at-a-time ranking, and expects deriving classes to instead provide `initial_score()` and `score_one()` implementations to define the scoring function used for each document. **Existing code that derived from `ranker` prior to this version of MeTA likely needs to be changed to instead derive from `ranking_function`.**

Additional functions have been added to `ranker_factory` to allow construction/loading of `language_model_ranker` subclasses (useful for the `kl_divergence_prf` implementation).

Regression tests have been added for rankers' MAP and NDCG scores. This adds a new dataset `cranfield` that contains non-binary relevance judgments to facilitate these new tests.

### Minor Changes
Adds the `util::transform_iterator` class and `util::make_transform_iterator` function for providing iterators that transform their output according to a unary function.

Adds a util::make_fixed_heap helper function to simplify the declaration of util::fixed_heap classes with lambda function comparators.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ doc/
 data/ceeaus
 data/breast-cancer
 data/housing
+data/cranfield
 biicode.conf
 bii/
 bin/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,13 +28,15 @@
   The EM algorithm used for the two-component mixture model is provided as
   the `index::feedback::unigram_mixture` free function and returns the
   feedback model.
-- **Breaking Change.** This change also breaks the `ranker` hierarchy into
-  one more level. At the top we have `ranker`, which has a pure virtual
-  function `rank()` that can be overridden to provide entirely custom
-  ranking behavior, This is the class the KL-divergence method derives
-  from, as we need to re-define what it means to rank documents (first
-  retrieving a feedback set, then ranking documents with respect to an
-  updated query).
+- Add the Rocchio algorithm (`rocchio`) for pseudo-relevance feedback in
+  the vector space model.
+- **Breaking Change.** To facilitate the above to changes, we have also
+  broken the `ranker` hierarchy into one more level. At the top we have
+  `ranker`, which has a pure virtual function `rank()` that can be
+  overridden to provide entirely custom ranking behavior, This is the class
+  the KL-divergence and Rocchio methods derive from, as we need to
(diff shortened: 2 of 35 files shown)


smassung:
LGTM.

Are you planning on adding Rocchio?

skystrife:
Just added it now.

smassung:
Usually there is not a reason to include the negative feedback docs because it hurts Rocchio performance. Is this the reason for not adding the gamma parameter? Or is it unclear what negative docs would be?

Also, out of curiosity, how does Rocchio performance compare to PFR on the Cranfield dataset? Based on the unit tests, it looks like Rocchio is slightly better when using the default params?

skystrife:
> Usually there is not a reason to include the negative feedback docs because it hurts Rocchio performance. Is this the reason for not adding the gamma parameter? Or is it unclear what negative docs would be?

Yes to both, really. It's unclear that this would matter much, and it's also unclear (to me) what to use as the negative documents.

> Also, out of curiosity, how does Rocchio performance compare to PFR on the Cranfield dataset? Based on the unit tests, it looks like Rocchio is slightly better when using the default params?

I assume you mean comparing Rocchio with the KL-divergence mixture model method? It looks like Rocchio is indeed doing better (at least on Cranfield; I haven't tried on a big TREC dataset yet), but I suspect this is more due to a bad default parameter for `dirichlet_prior` than the method itself being better. (I think `mu = 1000` is better than our current default of 2000.)

PR #175, opened 2016-12-09
Pseudo feedback framework

gress2:
adds:
feedback base class, feedback factory
rocchio feedback implementation
ide-dec-hi feedback implementation
document vector class for VSM representations of queries / documents

modified:
ranker base class to work with VSM queries

Looks like some feedback features were snuck into the develop branch while I was working on this so there are some conflicts..


gress2:
@skystrife I'm wondering if there's much value in this PR given what you've written for feedback

skystrife:
Sorry for the extended delay in getting back to you on this.

I think there is still value here, particularly in how you've separated out the feedback method from the rankers. I think it might be beneficial to be able to, for example, investigate which query terms were added by feedback, which in my current implementation is completely hidden away inside the ranker.

I would be open to a refactoring of the existing feedback code that extracts the feedback mechanism out into a separate class hierarchy like you have here, which will give users the option to either apply it manually (to see what exactly about a query is being modified) or use it as a "black box" ranker (which is what the current implementation does).

skystrife:
Closing for inactivity.

PR #177, opened 2017-02-07
Parallel cooccurrence counting

skystrife:
This PR parallelizes the cooccurrence counting application for the GloVe embedding method using the new `corpus::parallel_consume` functionality.

It also fixes the somewhat hilarious typo of `coocur` and replaces it with `cooccur` like I actually intended.


include/meta/embeddings/cooccurrence_counter.h
```diff
@@ -0,0 +1,225 @@
+/**
+ * @file coocurrence_counter.h
```
smassung: typo

include/meta/embeddings/cooccurrence_counter.h
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    uint64_t bytes_read() const
+    {
+        return sizeof(count_type) * idx_;
+    }
+
+    bool operator==(const memory_cooccur_iterator& other) const
+    {
+        return items_.empty() && other.items_.empty();
```
smassung: So it's never possible that `items_` is nonempty and equal to `other.items_`?

include/meta/util/multiway_merge.h
```diff
@@ -84,14 +85,21 @@ namespace util
  *     A unary function that is called once per every unique Record after
  *     merging.
  *
+ * - ProgresTrait:
```
smassung: typo

src/embeddings/tools/embedding_cooccur.cpp
```diff
@@ -0,0 +1,81 @@
+/**
+ * @file embedding_coocur.cpp
```
smassung: typo

src/embeddings/tools/glove.cpp
```diff
@@ -176,13 +176,13 @@ class glove_trainer
             throw glove_exception{"no vocabulary file found in " + prefix};
         }
 
-        if (!filesystem::file_exists(prefix + "/coocur.bin"))
+        if (!filesystem::file_exists(prefix + "/cooccur.bin"))
         {
             LOG(fatal)
                 << "Coocurrence matrix has not yet been generated, please "
```
smassung: typo

skystrife: No?

smassung: in the log text

include/meta/embeddings/cooccurrence_counter.h
skystrife: We only want `it1 == it2` when `it1` and `it2` are both the "end" iterator. This is only true when `items_` is empty.

src/embeddings/tools/glove.cpp
skystrife: Ah, now I see it.

include/meta/embeddings/cooccurrence_counter.h
smassung: Ah, I see.

smassung: Ah, I see.

src/embeddings/cooccurrence_counter.cpp
```diff
@@ -0,0 +1,349 @@
...
+                {
+                    history.clear();
+                }
+                else if (tok == "</s>" && break_on_tags_)
+                {
+                    continue;
+                }
+                else
+                {
+                    // ignore out-of-vocabulary words
```
smassung: (thought I left this comment already)

Add an option for dirty vs clean sampling? Not sure if it makes much difference.

skystrife: Levy and Goldberg have this to say:

> **Deleting  Rare  Words  (`del`)** While  it  is  common to ignore words that are rare in the training corpus, word2vec removes  these  tokens  from the corpus before creating context windows.   As with subsampling, this variation narrows the distance between tokens, inserting new word-context pairs that did not exist in the original corpus with the same window size. Though this variation may also  have  an  effect  on  performance,  preliminary experiments  showed  that  it  was  small,  and  we therefore do not investigate its effect in this paper.

Not sure if I care about adding it. It adds a bunch of if/else into this loop to handle the case where `tid` and/or `*it` is not actually a value in the vocabulary.

smassung:
Ready to merge? All looks good to me.

PR #181, opened 2017-03-07
make lda_models operate on datasets instead of forward_indexes

smassung:
PR for easier conversation.

Changes
- removed uses of `no_evict_cache`
- use `dataset_view` internally in `lda_scvb` to manage the minibatch shuffling

Things that could be added:
- making `doc_size` a function of `instance` or `dataset`; or just leaving as a helper function in `lda_model`
- constructor for `dataset` that takes a single `{inverted,forward}_index` parameter and automatically uses the whole document range

### DIFF
--- a/include/meta/topics/lda_cvb.h
+++ b/include/meta/topics/lda_cvb.h
@@ -42,7 +42,7 @@ class lda_cvb : public lda_model
      * @param beta The hyperparameter for the Dirichlet prior over
      *  \f$\theta\f$
      */
-    lda_cvb(learn::dataset docs, std::size_t num_topics, double alpha,
+    lda_cvb(const learn::dataset& docs, std::size_t num_topics, double alpha,
             double beta);
 
     /**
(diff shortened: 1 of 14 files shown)


skystrife:
Do we need to store a copy of the dataset internally?

smassung:
Oh, no. I just blindly converted from the index version!

skystrife:
Mm, I think my point is less "let's not copy it" and more "do we even need to store it at all"? We only require it for training, right?

smassung:
It is only required for training, but is needed in both the constructor and `run` functions for the models.

The way `topics` is set up is that the `lda_*` classes are the trainers. To read saved models, you use the `lda-topics` program, which reads the plaintext files output by the trainers (this only requires a `forward_index` to convert term IDs to their string representations).

Of course, this could be changed to make the `topics` namespace more like the rest of MeTA. That is, `lda_*` learns a model if one doesn't already exist and saves it to disk in serialized form. Then, we can programmatically read the saved models and interact with the various distributions over the topics API as opposed to calling `lda-topics` and printing output.

I think the way it's set up now makes sense for the "trainers" to hold a reference to the `dataset` that they are operating on. Alternatively, we could change the constructors to start the training (essentially having them call `run`), and then I don't think we'd need to store a `dataset` ref internally. We'd need to pass the number of iterations and convergence threshold to the constructor. Either is fine with me.

smassung:
Good to merge?

PR #184, opened 2017-06-07
Add SGNS to MeTA

mk3890:
Want to get some feedback. I imagine this will generate a lot of suggestions on improvements. 

@skystrife @shlemas


skystrife:
Build failures appear to be just CI related random breakage... I think it should build cleanly against the latest develop commits.

Starting to look at this now.

src/embeddings/tools/sgns.cpp
```diff
@@ -0,0 +1,573 @@
...
+            corpus::parallel_consume(
+                *docs, pool,
+                [&]() {
+                    return sgns_local_buffer{*stream};
+                },
+                [&](sgns_local_buffer& buffer, const corpus::document& doc) {
+                    unsigned seed = std::chrono::system_clock::now().time_since_epoch().count();
+                    sgns_net_vector neu1e(vector_size_, 0);        
+                    std::default_random_engine engine(seed);
+                    std::uniform_int_distribution<std::size_t> next_random;
```
skystrife: Actually, it might be good to move these into your `sgns_local_buffer` to avoid extra allocations/work occurring every time a new document is being processed.

src/embeddings/tools/sgns.cpp
```diff
@@ -0,0 +1,573 @@
...
+        if (!stream)
+        {
+            throw sgns_exception{"Failed to find an ngram-word analyzer configuration"};
+        }
+
+        parallel::thread_pool pool{num_threads_};
+
+        auto iterations_left = iterations_;
+
+        auto docs = corpus::make_corpus(cfg_);
```
shlemas: Did you try doing this at the top of the while loop so you don't have to add reset() and reset_metadata() methods?

include/meta/corpus/corpus.h
```diff
@@ -75,6 +75,16 @@ class corpus
...
 
     /**
+     * Reset the corpus to the beginning.
+     */
+    virtual void reset() = 0;
+
+    /**
+     * Skip the next n documents in this corpus (default = 1).
+     */
+    virtual void skip(uint64_t n = 1) = 0;
```
shlemas: Since this was our quick and hacky (and slow) way of parallelizing the corpus before, I think you can get rid of it and all the implementations in the various subclasses.

include/meta/corpus/corpus.h
```diff
@@ -113,6 +123,18 @@ class corpus
...
+     * Helper function to be used by deriving classes in implementing
+     * reset() to reset the metadata for the current document.
+     */
+    void reset_metadata();
+
+    /**
+     * Helper function to be used by deriving classes in implementing
+     * skip() to skip the metadata for the current document.
+     */
+    void skip_metadata(uint64_t n = 1);
```
shlemas: Can presumably get rid of this too.

src/embeddings/tools/sgns.cpp
```diff
@@ -0,0 +1,573 @@
...
+typedef std::deque<sgns_vocab_vector::size_type> sgns_window;
+
+// Represents the neural network that is trained.
+struct sgns_net
+{
+    sgns_net(std::size_t vocab_size, std::size_t layer1_size) :
+        syn0(vocab_size * layer1_size),
+        syn1neg(vocab_size * layer1_size, 0)
+    {
+        std::default_random_engine random; // TODO: Seed
```
shlemas: We left a TODO here.

src/embeddings/tools/sgns.cpp
```diff
@@ -0,0 +1,573 @@
...
+        num_threads_(embed_cfg.get_as<std::size_t>("num-threads").value_or(std::max(1u, std::thread::hardware_concurrency()))),
+        max_ram_(embed_cfg.get_as<std::size_t>("max-ram").value_or(4096) * 1024 * 1024),
+        subsample_threshold_(sgns_cfg.get_as<double>("subsample-threshold").value_or(1E-4)),
+        max_window_size_(sgns_cfg.get_as<std::size_t>("max-window-size").value_or(6)),
+        iterations_(sgns_cfg.get_as<uint64_t>("iterations").value_or(10)),
+        starting_learning_rate_((float)sgns_cfg.get_as<double>("learning-rate").value_or(0.025)),
+        negative_samples_(sgns_cfg.get_as<std::size_t>("negative-samples").value_or(20)),
+        binary_output_(sgns_cfg.get_as<bool>("binary").value_or(true)),
+        vocab_(load_vocab(prefix_ + "/vocab.bin")),
+        noise_dist_(create_unigram_noise_distribution(1E8)), // TODO: Configurable size?
```
shlemas: TODO here too.

src/embeddings/tools/sgns.cpp
```diff
@@ -0,0 +1,573 @@
...
+                                // The first sample here (n == 0) will use the real
+                                // target. The rest (n > 0) will be negative samples.
+                               
+                                for (std::size_t n = 0; n < negative_samples_ + 1; n++)
+                                {
+                                    int label = 1;
+                                    sgns_net_vector::size_type target = window[target_widx];
+
+                                    // TODO: Maybe we do the first sample explicity and eliminate the if logic
+                                    // with this loop? Wonder if it is worth trying?
```
shlemas: Another TODO. Could probably just delete this one.

src/embeddings/tools/sgns.cpp
```diff
@@ -0,0 +1,573 @@
...
+    }
+
+    // Create the noise distribution from which negative word samples are
+    // randomly drawn.
+    sgns_noise_distribution create_unigram_noise_distribution(sgns_noise_distribution::size_type size) const
+    {
+        printing::progress progress{" > Generating noise distribution: ", vocab_.vector.size()};
+
+        sgns_noise_distribution noise_dist(size);
+        const double power = 0.75; // TODO: Configurable?
```
shlemas: Another TODO.

src/embeddings/tools/sgns.cpp
```diff
@@ -0,0 +1,573 @@
...
+        progress.end();
+
+        LOG(info) << "Created smoothed unigram noise distribution of size "
+                  << noise_dist.size() << ENDLG;
+
+        return noise_dist;
+    }
+
+    // Save vectors in word2vec format.
+    void save_w2v_vectors() const
```
shlemas: Remove this function?

src/embeddings/tools/embedding_vocab.cpp
```diff
@@ -16,6 +16,28 @@
 
 using namespace meta;
 
+namespace
+{
+    const std::string SGNS  = "sgns";
+    const std::string GLOVE = "glove";
```
shlemas: This probably sucks now? Any better way?

src/embeddings/tools/sgns.cpp
```diff
@@ -0,0 +1,573 @@
...
+                        {
+                            // Don't process a target word as its own context word.
+                            // Max window size works here because it is window size + 1th index
+                            // and therefore the target word
+                            if (w != max_window_size_)
+                            {
+                                const int64_t context_widx = (int64_t)target_widx - (int64_t)max_window_size_ + (int64_t)w;
+
+                                if (context_widx < 0 || context_widx >= (int64_t)window.size())
+                                    continue;
```
skystrife: From line 190 to here is pretty hard for me to wrap my head around.

Is `window_size` actually the (symmetric) window size you're considering to the left and right of `target_widx`? Or is it like a starting position into the window of size `max_window_size_`? I'm pretty sure it is the latter, but I'm not entirely sure why. It seems like it would be clearer to me to do something like:

```cpp
// window_size: size of the symmetric window to the left and also to the right of the target word
// range is [1, max_window_size_]
const auto window_size = (next_random(engine) % max_window_size_) + 1;

// Sweep across the window
for (std::size_t context_widx = (target_widx - window_size); // TODO: do the right thing if window_size > target_widx
     context_widx < target_widx + window_size;
     ++context_widx)
{
    if (context_widx == target_widx)
        continue;
    // remaining code...
}
```

I might be missing something, though.

shlemas: (It's been a little while so bear with me.)

I recall trying better ways as you've described, running into issues (perhaps unrelated), and falling back to word2vec style which is what you see here. It's probably worth @mk3890 giving another shot at simplifying it. And it'll either work in which case hooray, or we'll relearn why we had to do this to begin with.

`window_size` is indeed a bad name for the variable, because like you pointed out, it's really leading to a window size of `max_window_size_ - window_size`.

src/embeddings/tools/embedding_vocab.cpp
```diff
@@ -16,6 +16,28 @@
...
+}
+
+hashing::probe_map<std::string, uint64_t> reduce_vocab(hashing::probe_map<std::string, uint64_t> vocab,
+                                                       unsigned int min_reduce)
+{
```
shlemas: I meant to comment on this one before too. What you said sounds familiar, and I think that is indeed why we did this.

include/meta/corpus/corpus.h
mk3890: Correct. Do you see any value in leaving in there or would you prefer to remove it?

src/embeddings/tools/sgns.cpp
```diff
@@ -0,0 +1,573 @@
...
+        sgns_local_buffer(const analyzers::token_stream& stream)
+            : stream_{stream.clone()}
+        { }
+
+        sgns_local_buffer(sgns_local_buffer&&) = default;
+
+    private:
+        friend class sgns_trainer;
+        std::unique_ptr<analyzers::token_stream> stream_;
+};
```
mk3890: The issue here was that the parallel_consume passes this into a function which causes it to attempt to copy the unqiue_ptr causing a compile error. That said, I think we'll move some things into it as mentioned below.

src/embeddings/tools/embedding_vocab.cpp
mk3890: I will investigate this to confirm the thought here.

src/embeddings/tools/sgns.cpp
```diff
@@ -0,0 +1,573 @@
...
+
+// Represents a word in the vocabulary.
+struct sgns_word
+{
+    std::string word;
+    uint64_t count;
+};
+
+typedef std::vector<sgns_word> sgns_vocab_vector;
```
mk3890: You learn something new everyday. We are deprived of C++11/14/17.

src/embeddings/tools/sgns.cpp
```diff
@@ -0,0 +1,573 @@
...
+        printing::progress progress{" > Training: ", (uint64_t)(starting_learning_rate_ * 100000.0)};
+
+        train(io_mutex, progress);
+
+        save_w2v_vectors();
+        save_meta_vectors();
+    }
+
+    // When the thread_pool goes out of scope, the threads will join. In other
+    // words, when this method returns, all threads have finished training.
```
mk3890: Will look into this.

src/embeddings/tools/sgns.cpp
```diff
@@ -0,0 +1,573 @@
...
+
+        while (iterations_left > 0)
+        {
+            std::uniform_int_distribution<std::size_t> next_random;
+            int64_t word_counter = 0;
+
+            corpus::parallel_consume(
+                *docs, pool,
+                [&]() {
+                    return sgns_local_buffer{*stream};
```
mk3890: Same as above. That was my initial cut but it returns a unique_ptr which it then tries to copy.

src/embeddings/tools/embedding_vocab.cpp
mk3890: The topic modeling portion of the library actually does this as well. That said, I'm open to suggestions.

I also could look at why we even did this. I recall it was due to some differences during vocabulary generation. It may not have even been worth it.

shlemas: Yeah, we are using it to call reduce_vocab when using SGNS but not with GloVe.

src/embeddings/tools/sgns.cpp
mk3890: Fixed.

src/embeddings/tools/sgns.cpp
mk3890: Fixed.

src/embeddings/tools/sgns.cpp
mk3890: Fixed.

src/embeddings/tools/sgns.cpp
mk3890: Moved this into another file and will turn it into a small utility.

include/meta/corpus/corpus.h
mk3890: Done.

include/meta/corpus/corpus.h
mk3890: Done.

include/meta/corpus/corpus.h
mk3890: Removed.

include/meta/corpus/corpus.h
mk3890: Removed.

include/meta/corpus/metadata_parser.h
```diff
@@ -40,6 +40,13 @@ class metadata_parser
      */
     std::vector<metadata::field> next();
 
+    void reset();
```
mk3890: Added.

include/meta/corpus/metadata_parser.h
```diff
@@ -40,6 +40,13 @@ class metadata_parser
...
+
+    /**
+     * Skips the metadata vector for the next n documents in the file (default = 1)
+     */
+    void skip(uint64_t n = 1);
```
mk3890: Removed.

src/embeddings/tools/sgns.cpp
```diff
@@ -0,0 +1,573 @@
+/**
+ * @file sgns.cpp
+ * @author Shawn LeMaster
+ * @author Matt Kelly
+ *
+ * Builds word embedding vectors using the Skip-Gram Negative Sampling method.
+ *
+ * @see https://code.google.com/archive/p/word2vec/
```
mk3890: I've added something but truly have no idea if I've done it in a good way or not...

src/embeddings/tools/sgns.cpp
mk3890: Fixed.

src/embeddings/tools/sgns.cpp
mk3890: Done. Seemed to provide a bit of a improvement.

src/embeddings/tools/sgns.cpp
mk3890: I did. This perhaps provide just a little benefit but the margin is slim enough I'm not totally convinced that change actually made the difference. I think I'll leave in until Chase profiles.

src/embeddings/tools/embedding_vocab.cpp
```diff
@@ -28,19 +50,30 @@ int main(int argc, char** argv)
...
     auto max_size = vocab_cfg->get_as<int64_t>("max-size")
                         .value_or(std::numeric_limits<int64_t>::max());
 
+    // Override if we have an SGNS max size set
+    max_size = vocab_cfg->get_as<int64_t>("max-size")
+                        .value_or(max_size);
+
+    auto sgns_cfg = vocab_cfg->get_table("sgns");
+    auto min_reduce = sgns_cfg ? sgns_cfg->get_as<unsigned int>("min-reduce").value_or(1) : 1;
+    auto min_reduce_threshold = sgns_cfg ? sgns_cfg->get_as<double>("min-reduce-threshold").value_or(0.7) : 0.7;
```
mk3890: Ran clang-format so this should be good now.

src/embeddings/tools/sgns.cpp
mk3890: Extracted a few functions to help with this.

src/embeddings/tools/sgns.cpp
```diff
@@ -0,0 +1,573 @@
...
+
+        fclose(file);
+    }
+
+    // Save vectors in MeTA format.
+    void save_meta_vectors() const
+    {
+        // Target embeddings.
+        {
+            std::ofstream output { prefix_ + "/embeddings.target.bin", std::ios::binary };
```
mk3890: Ran clang-format so these issue should be gone.

src/embeddings/tools/sgns.cpp
mk3890: I actually made this change as part of a refactor to extract some of this core loop into functions. I think it may have marginally helped.

src/embeddings/tools/sgns.cpp
mk3890: Done.

src/embeddings/tools/sgns.cpp
```diff
@@ -0,0 +1,573 @@
...
+
+        progress.end();
+
+        LOG(info) << "Created smoothed unigram noise distribution of size "
+                  << noise_dist.size() << ENDLG;
+
+        return noise_dist;
+    }
+
+    // Save vectors in word2vec format.
```
mk3890: Done.

include/meta/embeddings/word_embeddings.h
```diff
@@ -108,6 +123,12 @@ class word_embeddings
 
     /// The embeddings matrix
     util::aligned_vector<double> embeddings_;
+
+     /// The target vectors matrix
+    util::aligned_vector<double> target_vectors_;
+
+    /// The context vectors matrix
+    util::aligned_vector<double> context_vectors_;
```
skystrife: This triples the memory requirements for loading embeddings in the common case (target + context mode). I'm not sure I understand why this was added.

To access just target vectors, use the first constructor for `word_embeddings` with the target vector file as the second input stream. To access just the context vectors, use the first constructor of `word_embeddings` with the context vector file as the second input stream.

Is there something I'm missing?

src/embeddings/tools/meta_to_word2vec.cpp
```diff
@@ -0,0 +1,104 @@
...
+        throw word_embeddings_exception{
+            "missing prefix key in configuration file"};
+
+    auto binary_output = sgns_cfg->get_as<bool>("binary");
+    if (!prefix)
+        throw word_embeddings_exception{
+            "missing prefix key in configuration file"};
+
+    save_w2v_vectors(*binary_output, *prefix, load_embeddings(*embed_cfg));
+}
```
skystrife: `git` is complaining here about no trailing newline.

include/meta/embeddings/word_embeddings.h
mk3890: Nope. I made this refactor without realizing the mode switch was in there. 

Discovered that will debugging and forgot to remove it. Should be gone now.

src/embeddings/tools/meta_to_word2vec.cpp
mk3890: Beyond Compare is showing me a newline there, unless git is stripping it off during the commit.

mk3890:
@skystrife I think at this point I'm going to recommend we back burner the remaining 3 issues.

1. Creation of a temporary probe_map - I believe you are correct and when you push the alternative probe_map we can remove that and use erase()

The next two I'd like to take a look at but they are non-trivial and I'd want to do some serious testing to prove they are working. I started to do this with the 3rd issue but was having trouble getting deterministic results (even with 1 thread and the number generators seeded statically). I'd prefer to go with the current implementation for now and come back around if we have time at the end. That said, if you disagree I can keep digging on these things.

2. Usage of deque - I was doing some reading and saw articles that suggested deque was both basically as fast and those that show it was tremendously slower. I'm guessing it depends on the implementation/compiler. I definitely like how intuitive it is now, but like you said probably worth benchmarking.

3. Loop structure - I naively threw this in and it runs -- but it doesn't give us exactly the same result. From the prior work Shawn and I recall the difference being very nuanced so it would take some investigation. The result over wiki4gb are good, but they are distinctly different. It also shows an improbable speedup of something like 20%+ percent which makes me thing it isn't quite working correctly.

src/embeddings/tools/embedding_vocab.cpp
skystrife: It doesn't look like it on my end. Can you please re-run and commit that?

src/embeddings/tools/sgns.cpp
```diff
@@ -0,0 +1,607 @@
...
+
+                if (subsample_threshold_ > 0)
+                {
+                    const auto count = vocab_.vector[i->value()].count;
+                    const float ran
+                        = ((float)sqrt(count / (float)(subsample_threshold_
+                                                       * vocab_.total_count))
+                           + 1)
+                          * (float)(subsample_threshold_ * vocab_.total_count)
+                          / count;
```
skystrife: Oh jeez, `clang-format` really mangled this. Can you break this up into multiple statements for clarity?

Nit: `std::sqrt`.

src/embeddings/tools/sgns.cpp
```diff
@@ -0,0 +1,607 @@
...
+
+                    // The first sample here will use the real
+                    // target. The rest will be negative samples.
+                    update_vectors(1, window[target_widx], buffer.neu1e, l1);
+                    for (std::size_t n = 0; n < negative_samples_; n++)
+                    {
+                        sgns_net_vector::size_type target
+                            = noise_dist_[(buffer.next_random(buffer.engine)
+                                           >> 16)
+                                          % noise_dist_.size()];
```
skystrife: Why are we discarding the upper 16 bits of the random number here?

Let's also use `random::bounded_rand(buffer.engine, noise_dist_.size())` here for clarity and correctness. The modulo here introduces bias, and we have a function for this already that we can use that isn't biased. Might as well use it. It should also simplify things as we don't have to carry around the `buffer.next_random` distribution anymore.

src/embeddings/tools/sgns.cpp
```diff
@@ -0,0 +1,607 @@
...
+                if (subsample_threshold_ > 0)
+                {
+                    const auto count = vocab_.vector[i->value()].count;
+                    const float ran
+                        = ((float)sqrt(count / (float)(subsample_threshold_
+                                                       * vocab_.total_count))
+                           + 1)
+                          * (float)(subsample_threshold_ * vocab_.total_count)
+                          / count;
+                    if ((ran < (next_random(engine) & 0xFFFF) / 65536.0f))
```
skystrife: What's this about? Can we use `std::uniform_real_distribution<float> dist{0.0f, 1.0f};` and `ran < dist(engine)` here instead?

src/embeddings/tools/sgns.cpp
```diff
@@ -0,0 +1,607 @@
...
+            || ((max_ram_ - bytes_used) < (1024 * 1024)))
+        {
+            throw sgns_exception{"RAM limit too restrictive"};
+        }
+
+        return vocab;
+    }
+
+    // Create the noise distribution from which negative word samples are
+    // randomly drawn.
```
skystrife: Can you increase the verbosity of this comment just a little bit? This doesn't return a distribution; it returns a vector of length `size` whose elements are present in proportion to their noise sampling probability. Samples can then be drawn by sampling an index into the vector uniformly at random.

(I'm also curious if this even matters, performance wise. But that's how they do it, so presumably there's an improvement...)

src/embeddings/tools/sgns.cpp
```diff
@@ -0,0 +1,607 @@
...
+                                      double power) const
+    {
+        printing::progress progress{" > Generating noise distribution: ",
+                                    vocab_.vector.size()};
+
+        sgns_noise_distribution noise_dist(size);
+
+        double normalizer = 0;
+        for (sgns_vocab_vector::size_type i = 0; i < vocab_.vector.size(); ++i)
+            normalizer += pow(vocab_.vector[i].count, power);
```
skystrife: Nit: `std::pow`

src/embeddings/tools/sgns.cpp
```diff
@@ -0,0 +1,607 @@
...
+                         printing::progress& progress)
+    {
+        word_count_actual_ += word_counter;
+
+        learning_rate_ = std::max(
+            starting_learning_rate_
+                * (1.0f
+                   - (float)word_count_actual_
+                         / (float)(iterations_ * vocab_.total_count + 1.0f)),
+            starting_learning_rate_ * 0.0001f);
```
skystrife: This update is a race condition. I think it also is in the original word2vec, but it's perhaps worth noting.

src/embeddings/tools/sgns.cpp
```diff
@@ -0,0 +1,607 @@
...
+
+        {
+            const auto new_progress = (uint64_t)(
+                (starting_learning_rate_ - learning_rate_) * 100000.0);
```
skystrife: How does the learning rate relate to progress through an iteration? It seems like the natural thing here would be `progress(word_count_actual_)` with the constructor being `printing::progress progress{/* whatever */, iterations_ * vocab_.total_count};`, right?

src/embeddings/tools/sgns.cpp
```diff
@@ -0,0 +1,607 @@
...
+            std::lock_guard<std::mutex> lock{io_mutex};
+            progress(new_progress);
+        }
+    }
+
+    // Grabs the next word from the document stream, performs subsampling,
+    // and adds the vocabulary index for the word to the context window.
+    bool add_next_index(sgns_window& window,
+                        std::unique_ptr<meta::analyzers::token_stream>& stream,
```
skystrife: Nit: Can you dereference beforehand and just pass an `analyzers::token_stream&` here?

config.toml
```diff
@@ -95,9 +95,23 @@ dev-sections = [22, 22]
 test-sections = [23, 23]
 
 [embeddings]
+algorithm = "sgns"
 prefix = "word-embeddings"
 filter = [{type = "icu-tokenizer", suppress-tags = true}, {type = "lowercase"}]
 vector-size = 50
+num-threads = 4
```
skystrife: Why should the default configuration we provide be set to use only four threads?

src/embeddings/tools/sgns.cpp
skystrife: Can we actually remove the `reset()` and `reset_metadata()` methods unless my profiling indicates they are actually needed as a performance improvement? They are only needed here, and many of the corpora don't support them, so it's confusing from a use perspective. A common use case is likely going to be to run embeddings on compressed corpora, since you want to have a lot of raw text to train with.

(It may end up being the case that we don't even operate on a corpus here eventually, as I suspect the repeated tokenization process may be what slows us down relative to the original word2vec.)

src/embeddings/tools/embedding_vocab.cpp
```diff
@@ -51,7 +84,15 @@ int main(int argc, char** argv)
             stream->set_content(analyzers::get_content(doc));
 
             while (*stream)
+            {
                 ++vocab[stream->next()];
+
+                if (algorithm == SGNS && (vocab.size() >= min_reduce_threshold * max_size))
```
mk3890: Thanks for taking the time to understand what was going on and provide in-depth explanation. 

I've taken this to mean all the way back before we made any changes and reverted this file.

src/embeddings/tools/sgns.cpp
mk3890: This was an artifact of our "it isn't working let's converge with word2vec until it does" phase. 

Fixed.

src/embeddings/tools/sgns.cpp
mk3890: Agreed. Added.

src/embeddings/tools/sgns.cpp
mk3890: My initial thought was that word2vec did it this way, but that doesn't appear to be the case and I'm having trouble thinking of another reason we did it this way. It may have simply been to monitor that alpha was descending in the way we expected. That certainly isn't as accurate as the simple method you've proposed and that has been showing up in my recent runs.

Changed and I fixed a race condition with word_counter that was causing it to finish early.

src/embeddings/tools/sgns.cpp
mk3890: It's more legible now.

src/embeddings/tools/sgns.cpp
mk3890: Again an artifact of our "it isn't working let's converge with word2vec until it does" phase. 

Fixed.

src/embeddings/tools/sgns.cpp
mk3890: I tried (by using your words).

config.toml
mk3890: Removed.

src/corpus/corpus.cpp
```diff
@@ -3,9 +3,9 @@
  * @author Sean Massung
  */
 
+#include "meta/corpus/corpus.h"
 #include "cpptoml.h"
 #include "meta/corpus/all.h"
-#include "meta/corpus/corpus.h"
```
mk3890: clang-format moved it for me.

src/embeddings/tools/sgns.cpp
```diff
@@ -0,0 +1,606 @@
...
+    }
+
+    // Update the progress indicator
+    void update_progress(int64_t word_counter, std::mutex& io_mutex,
+                         printing::progress& progress)
+    {
+        // Note: This progress update is subject to a race condition
+        // when running multiple threads however this race condition
+        // also exists in the word2vec and doesn't appear to
+        // have any impact.
```
mk3890: Tricked myself. Changed the comment before the progress update.

src/embeddings/tools/sgns.cpp
```diff
@@ -0,0 +1,606 @@
...
+                io::packed::write(output, (double)i);
+                progress(++words_processed);
+            }
+
+            // Write out the unk vector.
+            for (std::size_t i = 0; i < vector_size_; ++i)
+            {
+                io::packed::write(output, 0.0);
+                progress(++words_processed);
+            }
```
mk3890: Nope. At the time I wasn't sure if it would make an impact or not. I can add it if you think it will provide an improvement.

src/corpus/corpus.cpp
skystrife: That's weird. Why would it de-alphabetize the includes?

I'm going to override `clang-format`'s opinion here...

src/embeddings/tools/sgns.cpp
```diff
@@ -0,0 +1,606 @@
...
+            corpus::parallel_consume(
+                *docs, pool,
+                [&]() {
+                    return sgns_local_buffer{*stream, vector_size_};
+                },
+                [&](sgns_local_buffer& buffer, const corpus::document& doc) {
+                    process_document(buffer, doc, io_mutex, progress);
+                });
+
+            // update_progress(word_counter, io_mutex, progress);
```
mk3890: Yeah it is very accurate even without it.

src/corpus/corpus.cpp
mk3890: It puts the file's own header first. Believe the order is

System
Self
Others

All respectively alphabetized.

skystrife: Hm. Ok. [This](http://llvm.org/docs/CodingStandards.html#include-style) makes a convincing argument, so I guess we should just leave it.

mk3890:
@skystrife Just checking, are we waiting on Sean, or one of the items that is still open?

skystrife:
Merged via 09d58e71764c74b0a5e62a08b9684d58a4ff2794 and 09d58e71764c74b0a5e62a08b9684d58a4ff2794 (for cleaner history).

Thanks @mk3890 and @shlemas!

skystrife:
If either @mk3890 or @shlemas (or both) have time, it would be great if you could issue a PR against the website at [meta-toolkit/meta-toolkit.org](/meta-toolkit/meta-toolkit.org) to update the word embeddings documentation to include how to learn vectors with SGNS, and how to convert MeTA's vector format to SGNS format.

mk3890:
I'll take a cut at it for both features.

Issue #164, opened 2016-10-01
Modernize LDA model storage format


skystrife:
Model saving should support writing to streams instead of fixed files, and should use the binary format from `io::packed`.

PR #186, opened 2017-07-11
Topic Model API Refactor 

mk3890:
Helps to branch off develop instead of master. 

This should address #164 and provide a cleaner API for dealing with the results of a topic model. The lda-topics tool has been updated to use this new API.


src/topics/lda_scvb.cpp
```diff
@@ -184,5 +185,18 @@ double lda_scvb::compute_doc_topic_probability(learn::instance_id doc,
     return (doc_topic_count_.at(doc).at(topic) + alpha_)
            / (doc_sizes_.at(doc) + num_topics_ * alpha_);
 }
+
+stats::multinomial<topic_id> lda_scvb::topic_distrbution(doc_id doc) const
+{
+    // TODO: Replace the count vectors with a multinomial rather than creating
+    // it here
+    stats::multinomial<topic_id> result = stats::multinomial<topic_id>();
```
skystrife: Nit: Why not just `stats::multinomial<topic_id> result;`?

src/topics/lda_model.cpp
```diff
@@ -16,30 +17,23 @@ lda_model::lda_model(const learn::dataset& docs, std::size_t num_topics)
...
-        file << doc.id << "\t";
-        double sum = 0;
+        auto dist = topic_distrbution(doc_id{d.id});
         for (topic_id j{0}; j < num_topics_; ++j)
         {
-            double prob = compute_doc_topic_probability(doc.id, j);
-            if (prob > 0)
-                file << j << ":" << prob << "\t";
-            sum += prob;
+            io::packed::write(stream, dist.counts(j));
```
skystrife: Did you mean `dist.probability(j)`?

In fact, can't this just be `io::packed::write(stream, dist)` instead? `stats::multinomial<T>` implements the `io::packed` overloads.

This whole thing would then be something like:

```cpp
io::packed::write(stream, docs_.size());
for (const auto& d : docs_)
{
    io::packed::write(stream, topic_distribution(doc_id{d.id}));
}
```

src/topics/lda_model.cpp
```diff
@@ -16,30 +17,23 @@ lda_model::lda_model(const learn::dataset& docs, std::size_t num_topics)
...
         }
-        if (std::abs(sum - 1) > 1e-6)
-            throw std::runtime_error{"invalid probability distribution"};
-        file << "\n";
     }
 }
 
-void lda_model::save_topic_term_distributions(const std::string& filename) const
+void lda_model::save_topic_term_distributions(std::ostream& stream) const
```
skystrife: I hate that this function doesn't do what it says. Can we move the functionality of computing normalized term scores to the display utility `lda-topics` instead?

I want this function to actually save the *real* topic-term distributions, not modified ones. The fact that it doesn't right now is surprising and wrong.

@smassung Do you remember where the heck this re-weighting formula comes from anyway? There's no citation.

src/topics/lda_scvb.cpp
```diff
@@ -184,5 +185,18 @@ double lda_scvb::compute_doc_topic_probability(learn::instance_id doc,
...
+    for (topic_id tid{0}; tid < num_topics_; ++tid)
+    {
+        result.increment(tid, doc_topic_count_.at(doc).at(tid));
```
skystrife: This is not correct; you are forgetting the prior. Either add `alpha` here, or just use `compute_doc_topic_probability(doc, tid)` for readability.

src/topics/tools/lda_topics.cpp
```diff
@@ -1,76 +1,59 @@
...
+        std::cerr << "Usage: " << argv[0] << " config.toml" << std::endl;
+        return 1;
+    }
+
+    logging::set_cerr_logging();
+
+    auto config = cpptoml::parse_file(argv[1]);
+    auto index
+        = index::make_index<index::forward_index, caching::no_evict_cache>(
+            *config);
```
skystrife: Heh... This should just be `index::make_index<index::forward_index>(*config)`. Would you mind changing it for me?

include/meta/topics/topic_model.h
```diff
@@ -0,0 +1,122 @@
...
+#include "meta/util/string_view.h"
+
+namespace meta
+{
+namespace topics
+{
+
+struct term_prob
+{
+    std::size_t tid;
```
skystrife: Nit: `term_id tid`

include/meta/topics/topic_model.h
```diff
@@ -0,0 +1,122 @@
...
+    double probability;
+};
+
+struct topic_prob
+{
+    std::size_t tid;
```
skystrife: Nit: `topic_id tid`;

src/topics/CMakeLists.txt
```diff
@@ -6,7 +6,8 @@ add_library(meta-topics lda_cvb.cpp
                         lda_gibbs.cpp
                         lda_model.cpp
                         lda_scvb.cpp
-                        parallel_lda_gibbs.cpp)
+                        parallel_lda_gibbs.cpp
+						topic_model.cpp)
```
skystrife: Nit: Spaces (yes) vs Tabs (no)

src/topics/lda_model.cpp
```diff
@@ -4,6 +4,7 @@
  */
 
 #include "meta/topics/lda_model.h"
+#include <iostream>
```
skystrife: Is this for `std::ofstream`? That would be `#include <fstream>`. Also, I like to put system includes above the meta includes, but that's so nitpicky...

src/topics/lda_model.cpp
```diff
@@ -52,25 +46,28 @@ void lda_model::save_topic_term_distributions(const std::string& filename) const
...
-        file << "\n";
     }
 }
 
 void lda_model::save(const std::string& prefix) const
 {
-    save_doc_topic_distributions(prefix + ".theta");
-    save_topic_term_distributions(prefix + ".phi");
+    std::ofstream theta_file{prefix + ".theta", std::ios::binary};
+    std::ofstream phi_file{prefix + ".phi", std::ios::binary};
```
skystrife: Nit: Can we suffix these with ".bin" now that they are binary files?

src/topics/tools/lda_topics.cpp
```diff
@@ -1,76 +1,59 @@
...
+#include "meta/logging/logger.h"
+#include "meta/topics/topic_model.h"
 
 using namespace meta;
+using namespace meta::topics;
 
-int print_usage(const std::string& name)
+int print_topics(std::shared_ptr<index::cached_index<index::forward_index,
+                                                     caching::no_evict_cache>>
+                     idx,
```
skystrife: Nit: Just pass a `const index::forward_index&` and dereference at the callsite.

src/topics/topic_model.cpp
```diff
@@ -0,0 +1,142 @@
...
+        {
+            throw topic_model_exception{"topic term stream ended unexpectedly"};
+        }
+
+        term_progress(tid);
+        auto& vec = topic_term_probabilities_[tid];
+        vec.resize(num_words_);
+        std::generate(vec.begin(), vec.end(),
+                      [&]() { return io::packed::read<double>(phi); });
+    }
```
skystrife: Might want a `term_progress.end()` here, or braces around the progress + loop?

include/meta/topics/topic_model.h
```diff
@@ -0,0 +1,122 @@
...
+     * @param phi The stream to read the vectors from
+     */
+    topic_model(std::istream& theta, std::istream& phi);
+
+    /**
+     * @param topic_id The topic to use
+     * @param k The number of words to return
+     * @return the top k most probable words in the topic
+     */
+    std::vector<term_prob> top_k(term_id topic_id, std::size_t k = 10) const;
```
skystrife: Did you mean `topic_id tid`?

src/topics/topic_model.cpp
```diff
@@ -0,0 +1,142 @@
...
+        doc_progress(d);
+        auto& dist = doc_topic_probabilities_[d];
+        for (topic_id j{0}; j < num_topics_; ++j)
+        {
+            dist.increment(j, io::packed::read<double>(theta));
+        }
+    }
+}
+
+std::vector<term_prob> topic_model::top_k(term_id topic_id, std::size_t k) const
```
skystrife: `topic_id tid`

src/topics/tools/lda_topics.cpp
```diff
@@ -1,76 +1,59 @@
...
-int print_topics(const std::string& config_file, const std::string& filename,
-                 size_t num_words)
-{
-    auto config = cpptoml::parse_file(config_file);
-    auto idx = index::make_index<index::forward_index>(*config);
-
-    std::ifstream file{filename};
-    while (file)
+    auto num_topics = tm.num_topics();
+    for (term_id i{0}; i < num_topics; ++i)
```
skystrife: Term ids shouldn't loop over topics.

src/topics/tools/lda_topics.cpp
```diff
@@ -1,76 +1,59 @@
...
         {
-            std::string to_split;
-            stream >> to_split;
-            if (to_split.length() == 0)
-                continue;
-            size_t idx = to_split.find_first_of(':');
-            term_id term{std::stoul(to_split.substr(0, idx))};
-            double prob = std::stod(to_split.substr(idx + 1));
-            pairs.emplace(term, prob);
+            std::cout << idx->term_text(term_id{i.tid}) << " (" << i.tid
```
skystrife: If you chance the definition of `tid` in `term_prob`, you can get rid of the cast here.

src/topics/tools/lda_topics.cpp
```diff
@@ -1,76 +1,59 @@
...
-        auto pairs = util::make_fixed_heap<scored_term>(
-            num_words, [](const scored_term& a, const scored_term& b) {
-                return a.second > b.second;
-            });
+        std::cout << "Topic " << i << ":" << std::endl;
+        std::cout << "-----------------" << std::endl;
 
-        while (stream)
+        auto top_k = tm.top_k(i, 10);
+        for (auto& i : top_k)
```
skystrife: Are these mutable for any reason?

src/topics/tools/lda_topics.cpp
```diff
@@ -1,76 +1,59 @@
...
+
+    logging::set_cerr_logging();
+
+    auto config = cpptoml::parse_file(argv[1]);
+    auto index
+        = index::make_index<index::forward_index, caching::no_evict_cache>(
+            *config);
+    auto topic_model = topics::load_topic_model(*config);
+
+    print_topics(index, topic_model);
```
skystrife: Any particular reason you removed the ability to specify the number of words to print on the command line?

include/meta/topics/topic_model.h
```diff
@@ -0,0 +1,122 @@
...
+     * given document
+     */
+    stats::multinomial<topic_id> topic_distribution(doc_id doc) const;
+
+    /**
+     * @param topic_id The topic we are concerned with
+     * @param term_id The term we are concerned with
+     * @return The probability of the term for the given topic
+     */
+    term_prob term_probability(topic_id topic_id, term_id term_id) const;
```
mk3890: Naming is hard...

I'll move it to a double and do likewise for topic_probability()

include/meta/topics/topic_model.h
```diff
@@ -0,0 +1,122 @@
...
+
+    /**
+     * @param doc The document we are concerned with
+     * @param topic_id The topic we are concerned with
+     * @return The probability for the given topic
+     */
+    topic_prob topic_probability(doc_id doc, topic_id topic_id) const;
+
+    const std::size_t& num_topics() const;
```
mk3890: The habit runs deep.

src/topics/lda_scvb.cpp
mk3890: It seems like I do need to incorporate alpha, but this isn't setting the probability (multinomial doesn't provide an interface for doing so) which is what compute_doc_topic_probability() would get me, no?

src/topics/lda_model.cpp
mk3890: Nope. I was hoping to use the probabilities, but in order to reconstruct the distribution you can only do so with the counts so far as I can tell.

That said, I didn't notice that multinomial implemented those functions which is great and I've moved to that.

src/topics/lda_model.cpp
mk3890: I've moved this to lda_topics for now, but should we add it into topic_model perhaps a normalized_term_prob()?

src/topics/tools/lda_topics.cpp
mk3890: Just forgot to add it back in.

I've added it back in as the config parameter "display-topics" (possibly the worst name ever). If can also add back as a command line parameter if you really prefer that.

src/topics/lda_scvb.cpp
mk3890: @skystrife Was hoping it wouldn't collapse this one...maybe this will open it back up.

skystrife: It probably only matters if you intend on using `dist.counts(event)` later on and expect it to give you an integer. The probability values coming out of `dist.probability(event)` would be equivalent (it just returns `counts(event) / counts()` which would end up being `probability(event) / 1.0` if you use `increment()` with probabilities instead of counts.

The alternative would just be

```cpp
result.increment(tid, doc_topic_count_.at(doc).at(tid) + alpha_);
```

mk3890: Oh...I guess that makes sense, but it seems a little counter intuitive to put it in a state where it won't quite meet the API when we can just use the alternative and it will. If I'm understanding that correctly...

I'll just append alpha.

src/topics/lda_model.cpp
smassung: Found the formula from slide 19 here: https://courses.engr.illinois.edu/cs598jhm/sp2010/Slides/Lecture06.pdf

smassung: Seems to be taken from this paper by Blei: http://www.cs.columbia.edu/~blei/papers/BleiLafferty2009.pdf

skystrife:
Looks like the tests need a few file names updated.

src/topics/lda_model.cpp
skystrife: My suggestion here is to have `top_k` take a scorer parameter, which defines a

```cpp
double operator()(topic_id k, term_id v) const
```

This lets you either pass in a lambda or a functor to the function to define how you want terms to be scored with respect to a topic. So we could do something like

```cpp
auto best = model.top_k(topic, 10, [&](topic_id k, term_id v) {
    return model.term_probability(k, v);
});
```

to get the top-k using just raw probabilities, and something like

```cpp
topics::default_term_scorer scorer{model};
auto best = model.top_k(topic, 10, scorer);
```

to get things with the tfidf-like scorer. (Having it be a separate class allows us to cache the denominator for each vocab item, since it won't change as a function of k.)

I recommend implementing `default_term_scorer` by using log properties and `fastapprox::fastlog` for the logarithms.

```
score(k,v) = \phi_{k,v} (\log \phi_{k,v} - 1/K \sum_j \phi_{j,v})
```

(If anybody can come up with a better name than `default_term_scorer` I'm all ears...)

mk3890: OK, I've implemented this setup in general along with a default_term_scorer (more on that below) and a bl_term_scorer based on the Blei-Lafferty weighting function. I was able to prove it was working with the BL function.

As for the "default", please double check my implementation of your formula above as the results are very clearly wrong, but I'm having a hard time figuring out why. 

For the time being, I've left the bl_term_scorer in lda_topics.

skystrife: Where you have

```cpp
sum *= (1 / model.num_topics());
```

you probably want

```cpp
sum *= (1.0 / model.num_topics());
```

as I think what's happening may be `sum *= 0;` because of integer math. Hopefully that fixes it?

(I also like the `bl_term_scorer` name.)

include/meta/topics/topic_model.h
```diff
@@ -57,7 +57,8 @@ class topic_model
      * @param k The number of words to return
      * @return the top k most probable words in the topic
      */
-    std::vector<term_prob> top_k(topic_id tid, std::size_t k = 10) const;
+    template <typename T>
+    std::vector<term_prob> top_k(topic_id tid, std::size_t k, T score) const;
```
skystrife: Please take T by universal reference (`T&& score`). We don't want to copy the scorers that have a lot of internal state.

include/meta/topics/bl_term_scorer.h
```diff
@@ -0,0 +1,52 @@
...
+#include "meta/meta.h"
+#include "meta/topics/topic_model.h"
+
+namespace meta
+{
+namespace topics
+{
+
+/* Scores terms according to a tfidf-like weighting by Blei and Laffterty.
+ * @see  http://www.cs.columbia.edu/~blei/papers/BleiLafferty2009.pdf */
```
skystrife: Change to

```cpp
/** 
 * Scores terms according to...
 * @see ...
 */
```

so Doxygen picks it up properly.

include/meta/topics/default_term_scorer.h
```diff
@@ -0,0 +1,51 @@
...
+#include "meta/index/forward_index.h"
+#include "meta/meta.h"
+#include "meta/topics/topic_model.h"
+
+namespace meta
+{
+namespace topics
+{
+
+/* Scores terms in some way Chase can probably articulate and or source */
```
skystrife: It should be equivalent to the Blei and Lafferty scoring function, just using log properties to (hopefully?) be faster.

If we can get this working, I think it should be called `bl_term_scorer` since that's really what it is.

src/topics/lda_model.cpp
mk3890: Sure was. Good catch.

Does that formula have any name associated with it that we can use to create a better name than default?

Also, I left a source / description in the doxygen comment up to you for the time being.

include/meta/topics/bl_term_scorer.h
```diff
@@ -0,0 +1,52 @@
...
+
+    /**
+     * The denominators
+     */
+    std::vector<double> denoms_;
+};
+}
+}
+
+#endif
```
skystrife: Does your editor not place newlines at the end of files? Git is complaining here.

include/meta/topics/default_term_scorer.h
```diff
@@ -0,0 +1,51 @@
...
+
+    /**
+     * The sums
+     */
+    std::vector<double> sums_;
+};
+}
+}
+
+#endif
```
skystrife: Nit: another newline issue.

include/meta/topics/topic_model.h
```diff
@@ -57,7 +57,8 @@ class topic_model
      * @param k The number of words to return
      * @return the top k most probable words in the topic
      */
-    std::vector<term_prob> top_k(topic_id tid, std::size_t k = 10) const;
```
skystrife: Can we keep this and just have it be

```cpp
return top_k(tid, k, [this](topic_id t, term_id v) { return term_probability(t, v); });
```

?

src/topics/lda_model.cpp
skystrife: It's the Blei and Lafferty formula, just expanded using log properties to avoid `std::pow`  and potential underflow in the denominator calculation (multiplying lots of small probabilities together might underflow before you can pow and take the log). I would just rename it `bl_term_scorer` and remove the existing `bl_term_scorer` since they should be the same thing (up to the approximation for the log introduced by fastapprox).

include/meta/topics/topic_model.h
mk3890: Sure. I was looking for a decent way to do something like this but forgot.

include/meta/topics/bl_term_scorer.h
mk3890: Yeah so apparently Visual Studio doesn't do always place a new line at the end of the file. I also never see errors about it though. What do you do that causes you to see them?

include/meta/topics/default_term_scorer.h
mk3890: I've removed and renamed as suggested.

include/meta/topics/bl_term_scorer.h
```diff
@@ -0,0 +1,54 @@
...
+#include "meta/meta.h"
+#include "meta/topics/topic_model.h"
+
+namespace meta
+{
+namespace topics
+{
+
+/**
+ * Scores terms according to a tfidf - like weighting by Blei and Laffterty.
```
skystrife: s/Laffterty/Lafferty

include/meta/topics/bl_term_scorer.h
```diff
@@ -0,0 +1,54 @@
...
+ * @see  http://www.cs.columbia.edu/~blei/papers/BleiLafferty2009.pdf
```
skystrife: Nit: extra space

include/meta/topics/bl_term_scorer.h
```diff
@@ -0,0 +1,54 @@
...
+ */
+class bl_term_scorer
+{
+  public:
+    /**
+     * @param model The topic model to score
+     */
+    bl_term_scorer(topics::topic_model model);
```
skystrife: Sorry I didn't catch this earlier, but can we just store a `const topic_model&` to avoid having two copies of it hanging around while running `top_k` queries?

include/meta/topics/bl_term_scorer.h
skystrife: They show up in the Files changed tab within the PR itself on GitHub.

src/topics/topic_model.cpp
```diff
@@ -0,0 +1,136 @@
...
+{
+    return doc_topic_probabilities_[doc];
+}
+
+double topic_model::term_probability(topic_id top_id, term_id tid) const
+{
+
+    auto prob = topic_term_probabilities_[top_id][tid];
+
+    return prob;
```
skystrife: `return topic_term_probabilities_[top_id][tid];`

src/topics/topic_model.cpp
```diff
@@ -0,0 +1,136 @@
...
+        throw topic_model_exception{"Missing prefix key in configuration file"};
+    }
+
+    std::ifstream theta{*prefix + ".theta.bin", std::ios::binary};
+    std::ifstream phi{*prefix + ".phi.bin", std::ios::binary};
+
+    if (!theta)
+    {
+        throw topic_model_exception{
+            "missing document topic probabilities file:" + *prefix + ".theta"};
```
skystrife: `.theta.bin`

src/topics/topic_model.cpp
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    if (!phi)
+    {
+        throw topic_model_exception{
+            "missing topic term probabilities file:" + *prefix + ".phi"};
```
skystrife: `.phi.bin`

include/meta/topics/bl_term_scorer.h
```diff
@@ -20,16 +20,16 @@ namespace topics
...
+ * @see http://www.cs.columbia.edu/~blei/papers/BleiLafferty2009.pdf
  */
 class bl_term_scorer
 {
   public:
     /**
      * @param model The topic model to score
      */
-    bl_term_scorer(topics::topic_model model);
+    bl_term_scorer(topics::topic_model& model);
```
skystrife: `const topic_model&`

skystrife:
If you have time, it would be great if you could issue a PR against the website at [meta-toolkit/meta-toolkit.org](https://github.com/meta-toolkit/meta-toolkit.org) (again on the develop branch) to update the topic model documentation to include how to interact with inference results using the new API.

PR #221, opened 2020-12-13
Hotfix for Meta failing icu download url for metapy

ccasey645:
Add updated ICU download URL after ICU dependency's tgz file was migrated from Subversion repo to GitHub repo (source: http://site.icu-project.org/repository).

Purpose: Metapy's Python wheel install process runs the Meta repo's cmake build process. The Meta repo's cmake steps will fail to download the ICU dependency because the old URL for the ICU dependency does not have the Tar file anymore. Without this updated URL that has the ICU dependency Meta needs to build, no-one can make builds of Meta and by extension, the Metapy repo cannot make new builds for Python 3.8 and Python 3.9 which are currently needed (but missing because of this bug).


ccasey645:
Closing: for some reason lots of extra commits were included with this PR. I branched off current commit hash used by Metapy repo which must not be from the master branch.

